PR #388, opened 2018-07-10
DLBCL IMPACT subset

mandawilson:
DLBCL subset of Lymphoma Super Cohort

NOTE: Oncotree code DLBCL is not part of the new oncotree, it is replace by DLBCLNOS, so subset on DLBCLNOS.

TODO: 
* Study needs to be created in Mercurial in dmp directory (it is currently in private).
* Deal with extra/empty attributes?


sheridancbio:
Sorry for ignoring this PR for three years ... is this change still needed? Should we actually complete this change or was this accomplished through another PR?

ao508:
@n1zea144 Do we still need this?

PR #575, opened 2019-05-03
Remove email of invalid timeline records

averyniceday:
This behavior will be captured inside the validator instead

This SHOULD NOT be merged until the validator (from cbioportal) is updated and deployed.


sheridancbio:
Do we know whether this has been added to the validator? Is this pending work, or is this PR obsolete?

PR #658, opened 2019-12-12
fix bug when multiple restart fails occur

sheridancbio:
we were getting emails from the cron daemon such as this:
/data/portal-cron/scripts/import-portal-users.sh: line 37: [: cgds_gdac: binary operator expected
when both kubernetes deployments failed to restart. This fixes this bug by wrapping the list of failed redeployments in quotes to form a string before doing the empty-string test.

### DIFF
--- a/import-scripts/restart-portal-pods.sh
+++ b/import-scripts/restart-portal-pods.sh
@@ -29,7 +29,7 @@ if [ -z $portal_id ] ; then
     exit 1
 fi
 deployment_id=${portal_to_deployment_map[$portal_id]}
-if [ -z $deployment_id ] ; then
+if [ -z "$deployment_id" ] ; then
     echo "invalid portal_id : $portal_id"
     print_portal_id_values
 fi


import-scripts/dddmmm
```diff
@@ -0,0 +1,404 @@
+#!/bin/bash
```
ao508: @sheridancbio what is this file name?? lol

import-scripts/importer.properties
```diff
@@ -0,0 +1,7 @@
+#comment
+#p1=hi
+p1=p1value
+p2=p2value
+p3=p3value
+ p4 = p4value
```
ao508: whitespace

import-scripts/dddmmm
sheridancbio: this was not supposed to be part of the changeset ... it was a temporary file. I'll remove it.

import-scripts/importer.properties
sheridancbio: the whitespace was intentional ... this was a test file added to make sure that flanking whitespace was handled properly while reading property names. But I did not mean to check in this test file. I'll remove it.

sheridancbio:
Sorry ... I accidentally added 3 testing files to the PR .. the actual code change is a one-line change. I'll fix that now.

PR #662, opened 2020-01-06
update slack webhook for logging messages

ao508:


averyniceday:
addressed in PR #663

PR #678, opened 2020-02-21
add script for deletions of terminating pods

sheridancbio:


sheridancbio:
some small bugfixes done during final testing

PR #679, opened 2020-02-24
Update annotation to use new POST method

averyniceday:
- remove partitioning code - now handled inside annotator
- logging can be added by adding to log4j and setting log level for annotator

**GN annotation pipeline hash HAS NOT been updated yet. This will be updated after GN-annotation-pipeline is brought up to date with master and unit tests fixed**

### DIFF
--- a/cvr/pom.xml
+++ b/cvr/pom.xml
@@ -67,7 +67,7 @@
     <dependency>
       <groupId>com.github.genome-nexus</groupId>
       <artifactId>genome-nexus-annotation-pipeline</artifactId>
-      <version>7b23109af9</version>
+      <version>05288349d1</version>
     </dependency>
     <dependency>
       <groupId>javax.mail</groupId>
--- a/cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/mutation/CVRMutationDataReader.java
+++ b/cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/mutation/CVRMutationDataReader.java
@@ -178,24 +178,20 @@ public class CVRMutationDataReader implements ItemStreamReader<AnnotatedRecord> 
         annotateRecordsWithPOST(recordsToAnnotate);
     }
 
-    private List<AnnotatedRecord> annotateRecordsWithPOST(List<MutationRecord> records) throws Exception {
-        List<AnnotatedRecord> annotatedRecordsList = new ArrayList<>();
-        List<List<MutationRecord>> partitionedMutationRecordsList = cvrUtilities.partitionMutationRecordsListForPOST(records, postIntervalSize);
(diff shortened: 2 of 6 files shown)


cvr/pom.xml
```diff
@@ -65,9 +65,9 @@
     used in the pipelines importer always matches the version used by the Genome
     Nexus annotation pipeline. -->
     <dependency>
-      <groupId>com.github.genome-nexus</groupId>
+      <groupId>com.github.averyniceday</groupId>
```
ao508: reminder to update the `artifactId` before merge.

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/mutation/CVRMutationDataReader.java
```diff
@@ -178,24 +178,23 @@ public void handleLine(String line) {
         annotateRecordsWithPOST(recordsToAnnotate);
     }
 
+    // doesn't seem like this method needs to return anything, could be a void function
```
ao508: let's just remove this comment then and make the method return void instead of leaving it as is.

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/mutation/CVRUnfilteredMutationDataReader.java
```diff
@@ -173,23 +173,20 @@ public void handleLine(String line) {
...
-                logAnnotationProgress(++annotatedVariantsCount, totalVariantsToAnnotateCount, postIntervalSize);
-                mutationRecords.add(ar);
-                mutationMap.getOrDefault(ar.getTUMOR_SAMPLE_BARCODE(), new ArrayList()).add(ar);
-                additionalPropertyKeys.addAll(ar.getAdditionalProperties().keySet());
-                header.addAll(ar.getHeaderWithAdditionalFields());
-            }
-            annotatedRecordsList.addAll(annotatedRecords);
+        // annotate with GenomeNexusImpl annotator from genome nexus annotation pipeline
+        // records will be partitioned inside annotator client
+        // records which do not get a response back will automatically be defauled to an AnnotatedRecord(record)
```
ao508: typo @ `defauled`

cvr/pom.xml
sheridancbio: done

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/mutation/CVRMutationDataReader.java
sheridancbio: already done

PR #688, opened 2020-03-03
Daily DMP Data Fetching and Updating Workflow

ao508:

### DIFF
--- a/import-scripts/dmp-import-vars-functions.sh
+++ b/import-scripts/dmp-import-vars-functions.sh
@@ -319,7 +319,7 @@ function remove_raw_clinical_timeline_data_files {
     done
     # remove raw timeline files except cbio format timeline file
     for f in $STUDY_DIRECTORY/data_timeline*; do
-        if [ $f != *"data_timeline.txt"* ] ; then
+        if [[ $f != *"data_timeline.txt"* ]] ; then
             $GIT_BINARY rm -f $f
         fi
     done
--- a/docs/DMP-Updates-Workflow.md
+++ b/docs/DMP-Updates-Workflow.md
@@ -1,60 +1,11 @@
-#!/bin/bash
+# DMP Updates Workflow
 
-# localize global variables / jar names and functions
-source $PORTAL_HOME/scripts/dmp-import-vars-functions.sh
+## 0. Pre-Update Steps
+**********************************************************************
 
-## STATUS FLAGS
+**Remove existing tmp files from $MSK_DMP_TMPDIR**
 
-# Flags indicating whether a study can be updated
-IMPORT_STATUS_IMPACT=0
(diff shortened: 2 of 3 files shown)


inodb:
@ao508 this is really helpful! Thanks for doing this!

inodb:
@angelica Not sure if it's helpful for me to review (yet) but just a few minor notes as someone not familiar with the process:

- github state: what is that and how is it used?
- dmp updates workflow seems to contain a lot of bash code. How do we make sure the documentation doesn’t get out of sync with the actual scripts? I assume there is some master script that runs all this?

sheridancbio:
There are extensive blocks of code shown in the documentation. It seems to me that these will soon be out of date unless we keep maintaining the documentation as we make code updates.

@ao508 suggests that we will automate a process to apply the headings and formatting to the code in order to do automatic updating of this documentation. It seems to me that that it will be a challenge to develop/maintain such automated code update scripts too.

import-scripts/dmp-import-vars-functions.sh
```diff
@@ -319,7 +319,7 @@ function remove_raw_clinical_timeline_data_files {
     done
     # remove raw timeline files except cbio format timeline file
     for f in $STUDY_DIRECTORY/data_timeline*; do
-        if [ $f != *"data_timeline.txt"* ] ; then
+        if [[ $f != *"data_timeline.txt"* ]] ; then
```
sheridancbio: I don't understand this line ... the fix seems fine, but I don't get the string comparison to a de-globbed (wildcard expanded) list of filenames. But that is a pre-existing bug (in my opinion)

ao508:
@inodb @sheridancbio Future plans include making a script that will generate a formatted workflow document. As per our morning discussion, this doc is good for merge now as is given the future plans.

inodb:
Thanks @ao508 ! What about the github states? Does that mean step 1/2/3 or do we keep track of the state id somewhere in code/config? Maybe i missed it somewhere in the docs

Re docs generation: could work. Is the idea to parse comments from some bash script? It might also work to split up the files a bit more so you can e.g. add a link to a script in the docs which has more detailed comments at the top of the file. As a more general comment: there are ofc many tools to generate docs from code, but if we have to develop our own tooling it could also be indicative of the code becoming too complex and that we've outgrown the existing setup. There are e.g. workflow languages that were specifically developed to handle these complexities, e.g. nextflow provides workflow tracing out of the box: https://www.nextflow.io/docs/latest/tracing.html

ao508:
@inodb Thanks for the pointer! 

Yes, we plan to capture GitHub states as well. The formatting will not be as "pretty" but I'm sure whatever automated workflow doc solution we end up using will be sufficient for what we need.

PR #689, opened 2020-03-03
Add validation of clinical record count after fetch

sheridancbio:
- drop of more than 90% not allowed with validate flag set


sheridancbio:
I know there is one bug in here .. will fix in about 2 hours (count of new records should not just come from the "new clinical record writer"

averyniceday:
assuming known bug is resolved ^

sheridancbio:
This PR is currently known to not be completed / tested.

PR #705, opened 2020-04-03
insulate oncokb-annotator from dmp runs

sheridancbio:
- wait until dmp import run is complete before starting
- copy needed source files to a staging directory
- use a separate working directory rather than working in repo subdirectory
use variable names for file paths
repo cleanup not needed because repo is now unaltered

### DIFF
--- a/import-scripts/create_web_directory_index_file.sh
+++ b/import-scripts/create_web_directory_index_file.sh
@@ -28,7 +28,8 @@ function make_directory_file() {
     echo "<title>$TITLE_STRING</title>" >> $INDEX_FILENAME
     echo "</head>" >> $INDEX_FILENAME
     echo "<body>" >> $INDEX_FILENAME
-    echo "<h2>Available Downloads</h2>" >> $INDEX_FILENAME
+    echo "<h2>OncoKB Annotated MSK-IMPACT</h2>" >> $INDEX_FILENAME
+    echo "<h3>Available Downloads</h3>" >> $INDEX_FILENAME
     echo "<p>Control-click links and choose \"Save Link As...\"" >> $INDEX_FILENAME
     echo "<ul>" >> $INDEX_FILENAME
     ls $SOURCE_DIR > $FILELIST_FILENAME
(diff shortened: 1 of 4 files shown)


sheridancbio:
One thing I wonder ... should we wrap this script with flock protection so that it cannot interfere with itself?

ao508:
@sheridancbio I think it would be a nice preventative measure just in case

sheridancbio:
I will add the flock protection. While testing I also added a timestamp for "last updated" to the generated index.html download page.

import-scripts/create_web_directory_index_file.sh
```diff
@@ -38,6 +39,8 @@ function make_directory_file() {
         fi
     done < $FILELIST_FILENAME
     echo "</ul>" >> $INDEX_FILENAME
+    timestamp=$(date '+%Y_%m_%d %H:%M %Z')
+    echo "<p style=\"color:Teal;font-size:7px;\">(last updated: $timestamp)" >> $INDEX_FILENAME
```
mandawilson, line 43: Good.

import-scripts/oncokb-annotator.sh
```diff
@@ -1,174 +1,298 @@
...
+            ONCOKB_ANNOTATION_SUCCESS=0
+        fi
+    fi
+
+    # Update DMP repo before running annotations
+    if [ $ONCOKB_ANNOTATION_SUCCESS -eq 1 ] ; then
+        echo $(date)
+        echo "Updating the DMP source repository..."
+        fetch_updates_in_data_sources dmp
+        if [ ${#failed_data_source_fetches[*]} -ne 0 ] ; then
```
mandawilson, line 158: `failed_data_source_fetches` set in `set-data-source-environment-vars.sh` and will contain a list of all data source fetches which failed after calling fetch_updates_in_data_sources()

import-scripts/oncokb-annotator.sh
```diff
@@ -1,174 +1,298 @@
...
+        echo $(date)
+        echo "Beginning README generation..."
+        $PYTHON_BINARY $README_SCRIPT -o $ONCOKB_README_FILE -u $ONCOKB_URL
+        if [ $? -ne 0 ] ; then
+            echo "Failed to generate README, exiting..."
+            ONCOKB_ANNOTATION_SUCCESS=0
+        fi
+    fi
+
+    # Only commit if all steps succeeded
```
mandawilson, line 278: Only rsync if all steps succeeded (instead of commit?)

sheridancbio: right .. there is no modification of the repository here. We upload the annotated oncokb files to ramen as a community download.

PR #706, opened 2020-04-07
Genie support tools

ao508:
Init commit of `annotatoin_suite_wrapper.sh`

#TODO: 
- [x] Parse input args to bash script correctly instead of assuming default paths and output merged MAF filename.
- [x] add call to convert VCFs to MAFs
- [x] add call to standardize MAFs (normalize columns headers so that all MAFs have the same header)
- [x] add call to annotation pipeline jar
- [x] add call to MAF merging tool 

This wrapper and the scripts called here would be replacing the following scripts under `import-scripts/annotation_suite`:
- apply-vcf-tools-per-center.py
- merge-and-annotate.py
- merge_mafs.py (might keep this one instead of using the existing script from the TARGET pipeline, this is to be determined)


sheridancbio:
I've only briefly scanned the work here. One top level question : is the inclusion of vcf2vcf necessary? It seems like this script is not called in the wrapper or any other script.

ao508:
@sheridancbio Yes, vcf2vcf should have been removed, looks like I missed it when I was cleaning up. It's now removed.

sheridancbio:
If I understand the flow now, the function of the outer script is to do the processing for a single center and the output of that processing is an annotated maf for all samples from that center. After that is run for all centers, the set of output mafs (one per center) will be merged by running  merge_mafs.py on the set of output per-center mafs.
Because the final step of merging mafs across centers is not part of the functionality of the outer script I think the name "annotation_suite_wrapper.sh" is too general for the outer script. I suggest a name like "annotate_center.sh" or "annotate_variant_files_for_center.sh" or "generate_annotated_maf_for_center.sh". That way, if a future script is developed which will iterate through a number of center directories, calling this outer script per center, and then will merge the per-center mafs into a single final maf automatically (full end-to-end functionality with a single script run) **that** script could have a general name like "annotation_suite_wrapper.sh" (although in that case I would suggest a name like "annotate_variants_for_all_centers.sh")

Also.. were we going to use python for the outer script? We are now only calling other python scripts.

import-scripts/annotation_suite/annotation_suite_wrapper.sh
```diff
@@ -0,0 +1,128 @@
...
+    java ${JAVA_SSL_ARGS} -jar ${GENOME_NEXUS_ANNOTATOR_JAR} --filename ${input_file} --output-filename ${output_file} --isoform-override ${GENOME_NEXUS_ANNOTATOR_ISOFORM} -r
+    if [ $? -gt 0 ] ; then
+        echo -e "\n[ERROR] annotateMAF(), error encountered while running the genome nexus annotation pipeline"
+        exit 1
+    fi
+}
+
+function annotateStandardizedMAFs {
+    for f in ${PROCESSED_SUB_DIR_NAME}/* ; do
+        if [ -f ${f} ] ; then
```
sheridancbio, line 106: if we allow spaces in filenames, these functions (this one, the one above, and below) need quoting around filenames

import-scripts/annotation_suite/merge_mafs.py
```diff
@@ -0,0 +1,122 @@
...
+    """ Generates a merged MAF given a list of input MAFs. """
+    merged_header = merge_headers(input_mafs)
+    sequenced_samples = set()
+
+    rows_to_write = []
+    for input_maf in input_mafs:
+        print('Loading data from ' + input_maf)
+        with open(input_maf, "r") as maf:
+            header_processed = False
+            file_header = get_file_header(input_maf)
```
sheridancbio, line 61: we could check here whether the header was empty too

import-scripts/annotation_suite/merge_mafs.py
```diff
@@ -0,0 +1,122 @@
...
+            for line in maf:
+                # add comment processing later (sequenced samples header?)
```
sheridancbio, line 63: we should discuss/resolve this

import-scripts/annotation_suite/merge_mafs.py
```diff
@@ -0,0 +1,122 @@
...
+
+def merge_input_mafs(input_mafs, output_maf_filename):
+    """ Generates a merged MAF given a list of input MAFs. """
+    merged_header = merge_headers(input_mafs)
+    sequenced_samples = set()
+
+    rows_to_write = []
+    for input_maf in input_mafs:
+        print('Loading data from ' + input_maf)
+        with open(input_maf, "r") as maf:
```
sheridancbio, line 59: It feels wrong (maybe it is fine) to open the file here, keep an open file handle to the file and then call get_file_header() below ... which will **also** open the file and have an open file handle. In other words, this script will have two simultaneous open file handles to the same file. I wonder if some operating systems will prevent that from working correctly. They are both "read" mode handles, which **should** not interfere with each other in theory. But maybe it would be reasonable to move the call to get_file_header above this line?

import-scripts/annotation_suite/merge_mafs.py
```diff
@@ -0,0 +1,122 @@
...
+                    continue
+                if not header_processed:
+                    header_processed = True
+                    continue
+                # map row values to current header columns
+                mapped_row = dict(zip(file_header, list(map(lambda x: process_datum(x), line.split("\t")))))
+                sample_id = mapped_row[TUMOR_SAMPLE_BARCODE_COLUMN]
+
+                if sample_id:
+                    sequenced_samples.add(sample_id)
```
sheridancbio, line 74: ok, so the merge script at least will add a sequenced_samples header based on what it has seen in the data table. But this does not include sequenced_samples from an input maf header which do not have any variation records in the data table. Again, we should discuss the needs for this header as a team.
Maybe we want to make the output of the sequenced_samples header optional?

import-scripts/annotation_suite/merge_mafs.py
```diff
@@ -0,0 +1,122 @@
...
+                # full record for the merged MAF (if mapped_row does not contain a column in merged_header, it is blank)
+                normalized_row = list(map(lambda x: mapped_row.get(x, ""), merged_header))
```
sheridancbio, line 76: nice

import-scripts/annotation_suite/merge_mafs.py
```diff
@@ -0,0 +1,122 @@
...
+                rows_to_write.append("\t".join(normalized_row))
+
+    with open(output_maf_filename, "w") as output_maf:
+        output_maf.write("#sequenced_samples: %s" % (" ".join(sequenced_samples)))
```
sheridancbio, line 80: maybe make this optional (command line option)

import-scripts/annotation_suite/merge_mafs.py
```diff
@@ -0,0 +1,122 @@
...
+        output_maf.write("\n")
```
sheridancbio, line 81: can't the line terminator be included in the prior write string?

import-scripts/annotation_suite/merge_mafs.py
```diff
@@ -0,0 +1,122 @@
...
+    input_mafs_list = args.input_mafs_list
+    input_mafs_directory = args.input_mafs_directory
+    output_maf = args.output_maf
+
+    if (input_mafs_list and input_mafs_directory) or (not input_mafs_list and not input_mafs_directory):
+        print("Please choose only one of the following options when running script:  --input-mafs-list | --input-mafs-directory")
+        usage(parser)
+
+    if input_mafs_list:
+        input_mafs = input_mafs_list.split(",")
```
sheridancbio, line 115: maybe map a "trim()" call to each member of this list (in case a comma separated list contains spaces and is in quotes)

import-scripts/annotation_suite/merge_mafs.py
```diff
@@ -0,0 +1,122 @@
...
+    else:
+        input_mafs = []
+        for maf in os.listdir(input_mafs_directory):
+            input_mafs.append(os.path.join(input_mafs_directory, maf))
```
sheridancbio, line 119: a check for inode type (directory v. file) might be good here .. and only append objects of type file

import-scripts/annotation_suite/README.md
```diff
@@ -0,0 +1,37 @@
+# ANNOTATION SUITE FOR GENIE
+
+Usage:
+
+```
+bash import-scripts/annotation_suite/annotation_suite_wrapper.sh
+annotation_suite_wrapper.sh
```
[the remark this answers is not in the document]
ao508: oh yeah I'll fix that - I just ran the wrapper without any arguments to get the printed usage statement out to copy and paste here. thanks!

import-scripts/annotation_suite/annotation_suite_wrapper.sh
```diff
@@ -0,0 +1,128 @@
...
+        rm -f $data_dir/*
+    else
+        mkdir $data_dir
+    fi
+}
+
+function standardizeMutationFilesFromDirectory {
+    # processed files will be written to ${OUTPUT_DATA_DIRECTORY}/processed
+    echo -e "\t[INFO] standardizeMutationFilesFromDirectory(), standardized mutation files from ${INPUT_DATA_DIRECTORY} will be written to ${PROCESSED_SUB_DIR_NAME}"
+    python3 ${STANDARDIZE_MUTATION_DATA_SCRIPT} --input-directory ${INPUT_DATA_DIRECTORY} --output-directory ${PROCESSED_SUB_DIR_NAME} --center ${CENTER_NAME} --sequence-source ${SEQUENCE_SOURCE} --extensions ${FILE_EXTENSIONS_LIST}
```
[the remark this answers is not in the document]
ao508: I dont think this will be necessary - if it doesnt exist then it wont be able to run and would exit anyway. Also given that the motivation for building this is for GENIE and Tom specifically runs python3 also makes me believe that this type of check isn't really necessary but I understand.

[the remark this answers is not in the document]
ao508: Actually now that I recall we want to migrate all of our python scripts to python3 at some point also.

import-scripts/annotation_suite/annotation_suite_wrapper.sh
ao508: I think I've put quotes around everything now (will be in next commit)

import-scripts/annotation_suite/merge_mafs.py
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import argparse
+# Script for mergining MAFs
+# Creates one aggregate header taken from column names acorss MAFs
+# Writes out rows while mapping to header
+# NAs and equivalents replaced with ''
+# - sequenced samples header???
```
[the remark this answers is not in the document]
ao508: i dont think the sequenced samples header should be supported due to the size of the genie so I'm going to remove it from here

import-scripts/annotation_suite/merge_mafs.py
```diff
@@ -0,0 +1,122 @@
...
+        Assumes file is tab-delimited
+        Assumes header is first non-commented line
+    """
+    header = []
+    with open(data_file, "r") as header_source:
+        for line in header_source:
+            if not line.startswith("#"):
+                header = list(map(str.strip, line.rstrip().split("\t")))
+                break
+    return header
```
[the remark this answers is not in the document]
ao508: in the event that a header does not exist that would imply that the file we are reading is actually empty and there wouldnt be anything to process

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1391 @@
...
+def print_warning(message):
+    print(message, ERROR_FILE)
+
+def resolve_variant_type(data, ref_allele, tum_seq_allele):
+    """ Resolve variant type.  """
+    for column in VARIANT_TYPE_COLUMNS:
+        if column in data.keys():
+            variant_type = process_datum(data[column].upper())
+            break
+    if variant_type == "1":
```
sheridancbio, line 268: are integers something we have seen before in the variant_type column in MAFs from genie?

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1391 @@
...
+        return "SNP"
+    elif variant_type == "2":
+        return "INS"
+    elif variant_type == "3":
+        return "DEL"
+
+    # if variant type is empty or not valid then try
+    # resolving it based on ref allele and tum seq allele values
+    if not variant_type in VALID_VARIANT_TYPES:
+        if len(ref_allele) > len(tum_seq_allele):
```
sheridancbio, line 278: I'm not familiar with this logic. One of the types for GenomicVariant in genome nexus is INDEL - my understanding of INDEL is an event where some of the reference genome has been deleted and in the place of the deletion there are some number of new nucleotides inserted. For INDEL, the reference genome allele length could be shorter than the variant allele length. Do we not allow INDEL events in our MAF files?
Looking here : https://docs.gdc.cancer.gov/Data/File_Formats/MAF_Format/ it does not seem to allow INDEL. There is a type called "Consolidated" but that appears to be a term used for rescinded records.

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1391 @@
...
+                variant_type = "ONP"
+
+    if variant_type == "":
+        message = "Could not salvage variant type from alleles [ ref allele = %s , tumor allele = %s ] " % (ref_allele, tum_seq_allele)
+        print_warning(message)
+
+    # if variant type still not in VALID_VARIANT_TYPES then must be SUB/SNV
+    if variant_type == "SNV":
+        variant_type = "SNP"
+    elif variant_type.upper() == "SUB":
```
sheridancbio, line 299: If SUB is a synonym for "Substitution" then I wonder if this used to be a code representing an indel. Trying to understand the ONP definition, the MAF_Format document above implies (in my opinion) that the reference allele length will match the variant allele length. It says "ONP  is analogous to TNP but for consecutive runs of four or more". So if SNP is 1 for 1, and DNP is 2 for 2, and TNP is 3 for 3, then ONP (by analogy) is N for N. This seems to not leave any mechanism for representing the deletion of 10 NT (such as ACGGCGGCAA) at a position and the insertion of 3 NT (such as TTT) in the same place. If SUB used to be a code which captured indels and then it was dropped from the MAF file standard, then perhaps we should not allow it here if the reference allele length does not equal the variant allele length

ao508: line 282 already does the check to determine whether the lengths of the reference and variant alleles match

import-scripts/annotation_suite/standardize_mutation_data.py
ao508: @sheridancbio this is parsing information from the input file which is not always in the standard MAF format with the standard/accepted values defined by the TCGA project.

import-scripts/annotation_suite/README.md
```diff
@@ -0,0 +1,37 @@
+# ANNOTATION SUITE FOR GENIE
```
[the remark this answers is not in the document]
ao508: @averyniceday had named it "annotation_suite" but I'm also a bit partial to it. I dont think it really matters significantly though if it's "suite" or "tool" though.

import-scripts/annotation_suite/merge_mafs.py
ao508: I'm following the same pattern we've used in our other scripts

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1391 @@
...
+}
+
+# KEY COLUMN NAMES
+TUMOR_SEQ_ALLELE1_COLUMNS = ["Tumor_Seq_Allele1", "TumorSeq_Allele1"]
+TUMOR_SEQ_ALLELE2_COLUMNS = ["Tumor_Seq_Allele2", "TumorSeq_Allele2"]
+MUTATED_TO_ALLELE_COLUMN = "mutated_to_allele"
+VARIANT_TYPE_COLUMNS = ["Variant_Type", "VariantType", "mut_type", "mutation_type"]
+VARIANT_CLASSIFICATION_COLUMNS = ["Variant_Classification", "class", "Transcript architecture around variant"]
+START_POSITION_COLUMNS = ["Start_Position", "Start_position", "POS", "chromosome_start"]
+END_POSITION_COLUMNS = ["End_Position", "End_position", "chromosome_end"]
```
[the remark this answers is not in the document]
ao508: no it's not from vcf2maf.. this script was written to handle more than just what we expect from GENIE since it was written while I was working on TARGET/CGCI which all submitted various MAFs generated from different pipelines.

import-scripts/annotation_suite/standardize_mutation_data.py
ao508: Same comment as before - this script was written for handling more than just what we expect from GENIE (or Tom expects from the individual centers)

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1391 @@
...
+def get_file_header(filename):
+    """ Returns the file header as a list. """
+    header = []
+    with open(filename, "rU") as data_file:
+        for line in data_file.readlines():
+            if line.startswith("#"):
+                continue
+            header = list(map(str.strip, line.split("\t")))
+            break
+    return header
```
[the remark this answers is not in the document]
ao508: will add

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1399 @@
...
+
+    # 8. cgpPINDEL: handle VCF lines from cgpPindel, where ALT depth and total depth are in PP:NP:PR:NR
+    elif is_cgppindel_vcf(vcf_format_data_keys):
+        # reference allele depth and depths for any other ALT alleles must be left undefined
+        allele_depth_values[variant_allele_idx] = str(int(mapped_sample_format_data["PP"]) + int(mapped_sample_format_data["NP"]))
+        mapped_sample_format_data["DP"] = str(int(mapped_sample_format_data["PR"]) + int(mapped_sample_format_data["NR"]))
+
+    # 9. VCF ALT allele fractions: handle VCF lines with ALT allele fraction in FA, which needs to be multiplied by DP to get AD
+    elif is_alt_allele_fraction_vcf(vcf_format_data_keys) and mapped_sample_format_data["DP"] != ".":
+        allele_depth_values[variant_allele_idx] = "%.0f" % (int(mapped_sample_format_data["FA"]) + int(mapped_sample_format_data["DP"]))
```
sheridancbio, line 1089: This should be product of fraction alt allele times total depth (* in place of +)

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1399 @@
...
+        - 0 = reference allele (i.e., what is in the 'REF' field)
+        - 1 = the first allele listed in 'ALT'
+        - 2 = the second allele listed in 'ALT'
+    """
+
+    variant_allele_idx = []
+    # if genotype information is available and a call was made for the sample then
+    # choose the first non-REF allele seen in the sample genotype
+    if "GT" in tumor_sample_format_data.keys() and not tumor_sample_format_data["GT"] in ["", ".", "./."]:
+        tumor_sample_genotype_info = tumor_sample_format_data["GT"].split("/")
```
sheridancbio, line 929: The split also needs to be done for vertical bar .. I'm not sure what the distinction is between a vertical bar separator or a slash separator, but both are present in the example MAF from the reference documentation

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+
+def is_valid_vcf_maf_record(maf_record):
+    return not ((maf_record["Reference_Allele"] == "N" or maf_record["Reference_Allele"] == "") and maf_record["Tumor_Seq_Allele2"] == "")
+
+def get_vcf_file_header(filename):
+    vcf_file_header = []
+    with open(filename, "r") as vcf_file:
+        for line in vcf_file.readlines():
+            if line.startswith("#CHROM"):
+                vcf_file_header = list(map(str.strip, line.replace("#", "").split("\t")))
```
sheridancbio, line 1277: for safety, I would change to line.replace("#", "", 1) so that only the first number sign is removed. If the header has additional number signs (such as an illegal trailing comment) we should let that generate errors rather than possibly interpreting a trailing comment as a sample id. Maybe we should do:
vcf_file_header_line = line.replace("#", "", 1)
if "#" in vcf_file_header_line:
   error message and exit
vcf_file_header = list(map(str.strip, vcf_file_header_line.split("\t"))

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+        sys.exit(2)
+    return vcf_file_header
+
+def extract_vcf_data_from_file(filename, center_name, sequence_source):
+    """ Load data from a VCF file. """
+    rejected_vars_by_varclass = 0
+    rejected_vars_by_mutstatus = 0
+    rejected_vars_by_inc_data = 0
+
+    (sample_id, tumor_sample_data_col, matched_normal_sample_id) = get_vcf_sample_and_normal_ids(filename)
```
sheridancbio, line 1290: get_vcf_sample_and_normal_ids() is a call into the unreviewed block of code from pre-existing target processing tools (~308-971)
however, I did notice in that code that the column names "TUMOR" and "NORMAL" are hardcoded there as the placeholders for identifying the samples when the sample id is not in the file (presumably taken from the filename instead). This should also be part of the README ... that these keywords are required when the sample_id is not from the filename.

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+    is_germline_data = ("germline" in filename)
+    data_file = open(filename, "r")
+    data_reader = [line for line in data_file.readlines() if not line.startswith("#")] # ignore commented lines
+    maf_data = []
+    vcf_data_errors = []
+    print("Loading data from file: %s" % (filename))
+    for line in data_reader:
+        vcf_file_header = get_vcf_file_header(filename)
```
sheridancbio, line 1298: Doesn't that mean that when processing each line of the data file, the file is opened again and again just to determine the header?
Let's move this outside the loop.

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+        vcf_data = dict(zip(vcf_file_header, list(map(str.strip, line.split("\t")))))
```
sheridancbio, line 1299: hopefully we have a rectangular matrix in the vcf file. should we validate that? (it could be done in a separate function before the data processing begins)

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+        # do some additional processing on the VCF columns "INFO", "FORMAT", the identified tumor_sample_data_col, and the NORMAL column (if NORMAL is present)
+        extract_vcf_format_info_data(vcf_data, tumor_sample_data_col, matched_normal_sample_id)
```
sheridancbio, line 1301: This is a call into the unreviewed block of code from pre-existing target processing tools (~308-971)

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+        maf_record = create_maf_record_from_vcf(sample_id, center_name, sequence_source, vcf_data, is_germline_data, matched_normal_sample_id, tumor_sample_data_col)
+        if not is_valid_vcf_maf_record(maf_record):
+            vcf_data_errors.append(vcf_data)
+        maf_data.append(maf_record)
+    data_file.close()
+    print_data_loading_summary(filename, len(maf_data), rejected_vars_by_varclass, rejected_vars_by_mutstatus, rejected_vars_by_inc_data, vcf_data_errors)
+    return maf_data
+
```
sheridancbio, line 1309: drop extra blank line here

import-scripts/annotation_suite/annotation_suite_wrapper.sh
```diff
@@ -0,0 +1,128 @@
...
+fi
+
+PROCESSED_SUB_DIR_NAME="${OUTPUT_DATA_DIRECTORY}/processed"
+ANNOTATED_SUB_DIR_NAME="${OUTPUT_DATA_DIRECTORY}/annotated"
+FILE_EXTENSIONS_LIST="vcf,maf,txt" # text files are treated as MAFs to handle names like data_mutations_extended.txt
+
+STANDARDIZE_MUTATION_DATA_SCRIPT=${ANNOTATION_SUITE_SCRIPTS_HOME}/standardize_mutation_data.py
+GENOME_NEXUS_ANNOTATOR_JAR=${ANNOTATION_SUITE_SCRIPTS_HOME}/annotator.jar
+MERGE_MAFS_SCRIPT=${ANNOTATION_SUITE_SCRIPTS_HOME}/merge_mafs.py
+
```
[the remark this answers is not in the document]
ao508: adding the cert file as a requirement in the annotation suite scripts home instead of making it a required external env variable

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+    start_pos = resolve_start_position(vcf_data)
+    variant_type = ""
+    end_pos = ""
+    variant_class = ""
+
+    if (ref_allele != "N" or ref_allele != "") and alt_allele != "":
+        # indels from vcf need to be shifted by one nucleotide and start position needs to be incremented by one
+        if ref_allele[0] == alt_allele[0] and ref_allele != alt_allele:
+            # shift ref allele and alt allele by one nucleotide, set as "-" if len == 1
+            if len(ref_allele) == 1:
```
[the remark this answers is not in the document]
ao508: the second if statement is checking the length of the alt allele not the ref allele

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+
+        # resolve variant type, end position, and variant class
+        variant_type = resolve_vcf_variant_type(ref_allele, alt_allele)
+        end_pos = resolve_end_position(vcf_data, start_pos, variant_type, ref_allele)
+        variant_class = resolve_variant_classification(vcf_data, variant_type, ref_allele, alt_allele)
+
+    maf_data["Variant_Classification"] = variant_class
+    maf_data["Variant_Type"] = variant_type
+    maf_data["Reference_Allele"] = ref_allele
+    maf_data["Tumor_Seq_Allele1"] = ref_allele
```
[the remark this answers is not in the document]
ao508: this is our default behavior for mskimpact but i think we now what we do is: if  `Tumor_Seq_Allele1` exists and the value is not empty then use that, otherwise use the reference allele

I will update this to use this behavior

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+
+    for filename in os.listdir(input_directory):
+        extract_data = False
+        for ext in extensions:
+            if filename.endswith(ext) and "normal" not in filename:
+                extract_data = True
+                break
+        if extract_data:
+            output_filename = os.path.join(output_directory, filename + ".temp")
+            if has_vcf_data and ".vcf" in filename:
```
sheridancbio, line 1370: rename : has_vcf_data -> processing_vcf_data (because ".vcf" may have been in the extensions list but no .vcf files are present for a center)
Also, I wonder if this is necessary .. because in the setting of extract_data above, we look for valid matching extensions ... so I wonder how we might have set extract_data to true, but not be processing files of type vcf?

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+                maf_data = extract_vcf_data_from_file(os.path.join(input_directory, filename), center_name, sequence_source)
+            else:
+                maf_data = extract_maf_data_from_file(os.path.join(input_directory, filename), center_name, sequence_source)
+            write_standardized_mutation_file(maf_data, output_filename)
+
+def is_valid_record(maf_data):
```
sheridancbio, line 1376: this function is never called. If this function is not needed, maybe VALID_CHROMOSOMES is also not needed.

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+    if not maf_data["Chromosome"] in VALID_CHROMOSOMES:
+        return False
+    return True
+
+def usage():
+    print("python3 standardize_mutation_data.py --input-directory [path/to/input/directory] --output-directory [path/to/output/directory] --center [default name for center] --sequence-source [WGS | WXS] --extensions [comma-delimited list of extensions]")
```
sheridancbio, line 1382: do non-programmers know "delimited" as a term? Perhaps "comma-separated" is more readable to general users?

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+    sys.exit(1)
+
+def main():
+    parser = optparse.OptionParser()
+    parser.add_option("-i", "--input-directory", action = "store", dest = "inputdir", help = "input data directory")
+    parser.add_option("-o", "--output-directory", action = "store", dest = "outputdir", help = "output data directory")
+    parser.add_option("-c", "--center", action = "store", dest = "centername", help = "name of center (standard MAF field = 'Center')")
+    parser.add_option("-s", "--sequence-source", action = "store", dest = "seqsource", help = "Sequencing source (standard MAF field = 'Sequencing_Source'), e.g., WXS or WGS")
+    parser.add_option("-x", "--extensions", action = "store", dest = "extensions", help = "File extensions to look for (e.g., .vcf, .maf), comma-delimited list")
```
sheridancbio, line 1391: Maybe "File extensions to be processed"?
And "comma-separated" if that was changed above (previous comment)

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+
+    (options, args) = parser.parse_args()
+    input_directory = options.inputdir
+    center_name = options.centername
+    output_directory = options.outputdir
+    sequence_source = options.seqsource
+    extensions = options.extensions
+
+    if not input_directory or not output_directory or not center_name or not extensions:
+        print("Input directory, output filename, extension(s) to search for, and center name must be provided.")
```
sheridancbio, line 1401: "extension(s) to be processed"?
probably too verbose : "extension(s) of input filenames to be processed"

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+        usage()
+    if not sequence_source:
+        sequence_source = ""
+    else:
+        sequence_source = sequence_source.upper()
+    if not center_name:
+        center_name = "."
```
sheridancbio, line 1408: ok .. so if the user explicitly sets `--center=""` then we override that with a period?
Is it improper to have a maf with blank values in the "center" column?

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+
+    has_vcf_data = ("vcf" in extensions)
+    generate_maf_from_input_data(input_directory, output_directory, extensions.split(","), center_name, sequence_source, has_vcf_data)
```
sheridancbio, line 1411: do we need to trim the extensions here?  In case the user does `--extensions="vcf, maf, txt    "` ?

import-scripts/annotation_suite/standardize_mutation_data.py
ao508: the VCF file has a fixed header with a little flexibility in the tumor column name - I don't think this will be necessary after all

import-scripts/annotation_suite/standardize_mutation_data.py
ao508: thanks for catching that

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+
+
+    # resolve counts and variant allele data
+    resolve_vcf_counts_data(vcf_data, maf_data, matched_normal_sample_id, tumor_sample_data_col)
+    resolve_vcf_variant_allele_data(vcf_data, maf_data)
+    resolve_vcf_matched_normal_allele_data(vcf_data, maf_data, matched_normal_sample_id)
+    return maf_data
+
+def is_valid_vcf_maf_record(maf_record):
+    return not ((maf_record["Reference_Allele"] == "N" or maf_record["Reference_Allele"] == "") and maf_record["Tumor_Seq_Allele2"] == "")
```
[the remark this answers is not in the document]
ao508: this is based on how the tumor seq allele value is resolved which does not allow a value "N" to be set so we could expand the check for consistency but it also shouldnt be possible technically - doesnt hurt to expand though so I'll update this soon

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+    return maf_data
+
+
+def extract_maf_data_from_file(filename, center_name, sequence_source):
+    rejected_vars_by_varclass = 0
+    rejected_vars_by_mutstatus = 0
+    rejected_vars_by_inc_data = 0
+    records_loaded = 0
+
+    header = get_file_header(filename)
```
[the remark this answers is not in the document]
ao508: For GENIE in the past the addition of the sequenced samples header has been problematic when we've had to split the MAF into multiple chunks for importing. So now we just rely on the case list file instead of generating a case list from the sequenced samples header.

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+
+
+def generate_maf_from_input_data(input_directory, output_directory, extensions, center_name, sequence_source, has_vcf_data):
+    print("\nLoading data from input directory: %s" % (input_directory))
+    print("\n\tSearching for files with extensions: %s " % (", ".join(extensions)))
+
+    for filename in os.listdir(input_directory):
+        extract_data = False
+        for ext in extensions:
+            if filename.endswith(ext) and "normal" not in filename:
```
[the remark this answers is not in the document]
ao508: in python  the `os.listdir(<directory>)` method returns a list of file basenames in the directory so we should be fine.

import-scripts/annotation_suite/standardize_mutation_data.py
ao508: removing `has_vcf_data`

import-scripts/annotation_suite/standardize_mutation_data.py
ao508: deafult value of empty string is okay, the "." is left over from a small change I made to match the center name in the example output mafs from tom so I'm removing that now

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+
+    header = get_file_header(filename)
+    data_file = open(filename, "r")
+    data_reader = [line for line in data_file.readlines() if not line.startswith("#")][1:]
+    maf_data = []
+    print("\nLoading data from file: %s" % (filename))
+    for line in data_reader:
+        add_record = True
+        if list(map(str.strip, line.split("\t"))) == header:
+            print("Double header in file! Skipping this line")
```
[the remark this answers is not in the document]
sheridancbio: this handles some target source files

import-scripts/annotation_suite/standardize_mutation_data.py
```diff
@@ -0,0 +1,1414 @@
...
+def print_data_loading_summary(filename, records_loaded, rejected_vars_by_varclass, rejected_vars_by_mutstatus, rejected_vars_by_inc_data, vcf_data_errors):
+    print("\tTotal records loaded %s " % (records_loaded))
+    if rejected_vars_by_varclass > 0:
+        print("\tTotal records filtered by variant classification: %s" % (rejected_vars_by_varclass))
+    if rejected_vars_by_mutstatus > 0:
+        print("\tTotal records rejected due to LOH or Wildtype Mutation Status: %s" % (rejected_vars_by_mutstatus))
+    if rejected_vars_by_inc_data > 0:
+        print("\tTotal records rejected due to incomplete allele data: %s" % (rejected_vars_by_inc_data))
+    if len(vcf_data_errors) > 0:
+        print("\tNumber of VCR data errors: %s" % (len(vcf_data_errors)))
```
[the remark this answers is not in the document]
sheridancbio: maybe we should ask whether this would be helpful to users (the full parse error details) ?

import-scripts/annotation_suite/merge_mafs.py
sheridancbio: we have decided that we will not support the Sequenced Samples header --- we instead will base the sample list on a case list file constructed from clinical data

import-scripts/annotation_suite/merge_mafs.py
sheridancbio: for this purpose, sequenced sample headers are not supported. The main merge.py script can be used to merge MAF with the sequence sample header.

PR #652, opened 2019-12-11
Cvr annotate with posts

ao508:
These changes migrate the CVR annotation step from using GETs to POSTs to improve runtime performance.

Total runtime for the entire CVR pipeline on `mskimpact` with and without using POSTs:
- with GETs: ~2.5 hours
- with POSTs: ~30 min 

Updates in this PR result in a significant performance improvement. 

Other updates:
- Refactor CVRUtil methods to CVRUtilities

TODO BEFORE MERGING: 
- [ ] need to replace the `groupId` and `version` with latest master after `genome-nexus-annotation-pipeline` [PR 108](https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/108) is merged.

Signed-off-by: Angelica Ochoa <15623749+ao508@users.noreply.github.com>

PR #710, opened 2020-04-17
Reintroduce Re-annotation Flag

averyniceday:


cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/mutation/CVRMutationDataReader.java
```diff
@@ -137,7 +137,7 @@ private void loadMutationRecordsFromJson(CVRData cvrData) {
         }
         log.info("Loaded " + String.valueOf(recordsToAnnotate.size()) + " records from JSON");
         try {
-            annotateRecordsWithPOST(recordsToAnnotate);
+            annotateRecordsWithPOST(recordsToAnnotate, true);
```
averyniceday, line 140: https://github.com/knowledgesystems/cmo-pipelines/pull/652/files#diff-7258385239600627f9ec5efeef299ee3L146

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/mutation/CVRMutationDataReader.java
```diff
@@ -175,16 +175,16 @@ public void handleLine(String line) {
         }
         reader.close();
         log.info("Loaded " + String.valueOf(recordsToAnnotate.size()) + " records from MAF");
-        annotateRecordsWithPOST(recordsToAnnotate);
+        annotateRecordsWithPOST(recordsToAnnotate, forceAnnotation);
```
averyniceday, line 178: https://github.com/knowledgesystems/cmo-pipelines/pull/652/files#diff-7258385239600627f9ec5efeef299ee3L193

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/mutation/CVRMutationDataReader.java
averyniceday: Added this argument `reannotate` back in to differentiate between whether we need to "force annotation".

For records read from JSON we want to ALWAYS annotate everything. This original logic was lost -- refer to link above to see what was dropped.

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/mutation/CVRMutationDataReader.java
averyniceday: Added this argument `reannotate` back in to differentiate between whether we need to "force annotation".

For records read from JSON we only want to annotate "unannotated records" unless we "force annotation". This original logic was lost -- refer to link above to see what was dropped.

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/mutation/CVRMutationDataReader.java
```diff
@@ -175,16 +175,16 @@ public void handleLine(String line) {
...
     }
 
-    private void annotateRecordsWithPOST(List<MutationRecord> records) throws Exception {
+    private void annotateRecordsWithPOST(List<MutationRecord> records, boolean reannotate) throws Exception {
```
averyniceday, line 181: Explained above but just to clarify - at some point we lost the use of the force annotation flag. We are putting it back in. But we also need to make sure we can still differentiate between records read from JSON and records read from MAF.

cvr/pom.xml
```diff
@@ -67,7 +67,7 @@
     <dependency>
       <groupId>com.github.genome-nexus</groupId>
       <artifactId>genome-nexus-annotation-pipeline</artifactId>
-      <version>48474a5d86</version>
+      <version>164ff61f07761423c129571e5364085b01e9a84d</version>
```
ao508, line 70: this will just need to get updated once the GN annotation pipeline PR gets merged - I will give @sheridancbio a chance to review first though

sheridancbio:
The build is failing. I am trying to determine why.

PR #721, opened 2020-04-24
Add POST size argument to annotation pipeline call

ao508:


ao508:
Waiting for additional feedback from Tom in case there are additional changes we need to make.

PR #731, opened 2020-05-14
Update CVR fetcher calls for ARCHER, ACCESS cohorts

ao508:
Added the `-r` arg to allow the pipeline to automatically drop samples from these cohorts if they are not in the master list.

### DIFF
--- a/import-scripts/fetch-dmp-data-for-import.sh
+++ b/import-scripts/fetch-dmp-data-for-import.sh
@@ -419,7 +419,7 @@ if [ $IMPORT_STATUS_ARCHER -eq 0 ] ; then
     # fetch new/updated archer samples using CVR Web service (must come after git fetching).
     printTimeStampedDataProcessingStepMessage "CVR fetch for archer"
     # archer has -b option to block warnings for samples with zero variants (all samples will have zero variants)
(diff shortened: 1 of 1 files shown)


import-scripts/fetch-dmp-data-for-import.sh
```diff
@@ -419,7 +419,7 @@ if [ $IMPORT_STATUS_ARCHER -eq 0 ] ; then
     # fetch new/updated archer samples using CVR Web service (must come after git fetching).
     printTimeStampedDataProcessingStepMessage "CVR fetch for archer"
     # archer has -b option to block warnings for samples with zero variants (all samples will have zero variants)
-    $JAVA_BINARY $JAVA_CVR_FETCHER_ARGS -d $MSK_ARCHER_UNFILTERED_DATA_HOME -n data_clinical_mskarcher_data_clinical.txt -i mskarcher -s -b $CVR_TEST_MODE_ARGS
+    $JAVA_BINARY $JAVA_CVR_FETCHER_ARGS -d $MSK_ARCHER_UNFILTERED_DATA_HOME -n data_clinical_mskarcher_data_clinical.txt -i mskarcher -s -b -r 50 $CVR_TEST_MODE_ARGS
```
mandawilson, line 422: @ao508 Why is there a maximum number of samples that can be removed, and why is it 50?  I assume it is so that we don't accidentally delete more than we should, but what if we should delete more than 50 (like someone did an update of many samples all at once). Does the script print/email a warning if it would have removed more?  I guess it would then remove the remaining samples over the next day (or days).

ao508: I made it 50 to match the threshold we set for the heme cohort

PR #753, opened 2020-06-24
Expand resolved sample type values for CVR

ao508:

### DIFF
--- a/cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/model/staging/CVRClinicalRecord.java
+++ b/cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/model/staging/CVRClinicalRecord.java
@@ -80,7 +80,7 @@ public class CVRClinicalRecord {
         this.sampleId = metaData.getDmpSampleId();
         this.patientId = metaData.getDmpPatientId();
         this.cancerType = metaData.getTumorTypeName();
-        this.sampleType = this.resolveSampleType(metaData.getIsMetastasis());
+        this.sampleType = resolveSampleType(metaData.getIsMetastasis());
         this.sampleClass = (studyId.equals("mskaccess")) ? MSKACCESS_SAMPLE_CLASS : DEFAULT_SAMPLE_CLASS;
         this.metastaticSite = metaData.getMetastasisSite();
         this.primarySite = metaData.getPrimarySite();
@@ -321,14 +321,14 @@ public class CVRClinicalRecord {
     public void setAGE_AT_SEQ_REPORT(String ageAtSeqReport) {
         this.ageAtSeqReport = ageAtSeqReport;
     }
-    
+
     public String getARCHER() {
         return archer != null ? archer : "NO";
     }
 
     public void setARCHER(String archer) {
         this.archer = archer;
-    }    
+    }
 
     public String getCVR_TMB_COHORT_PERCENTILE() {
         return cvrTmbCohortPercentile;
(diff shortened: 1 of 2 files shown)


cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/model/staging/CVRClinicalRecord.java
```diff
@@ -370,11 +370,20 @@ public void setMSK_SLIDE_ID(String mskSlideID) {
...
-        if (isMetastasis != null)
-            return isMetastasis == 0 ? "Primary" : "Metastasis";
+    public String resolveSampleType(Integer isMetastasis) {
+        if (isMetastasis != null) {
+            switch(isMetastasis) {
+                case 0:
+                    return "Primary";
+                case 1:
+                    return "Metastasis";
+                case 127:
```
sheridancbio, line 380: This is different from the description on the card. Local Recurrence had code of 2, and Unknown had a code of 127. Theoretically any other code coming in is an unexpected condition (maybe an exception or a reason to substitute NA)?

cvr/src/test/java/org/cbioportal/cmo/pipelines/cvr/CVRClinicalRecordTest.java
```diff
@@ -0,0 +1,70 @@
...
+        String sampleType = record.resolveSampleType(null);
+        Assert.assertTrue(sampleType.isEmpty());
+    }
+
+    @Test
+    public void testSampleTypeNotNull() throws Exception {
+        CVRClinicalRecord record = new CVRClinicalRecord();
+        String primarySampleType = record.resolveSampleType(0);
+        String metastasisSampleType = record.resolveSampleType(1);
+        String localRecurrenceSampleType = record.resolveSampleType(127);
```
sheridancbio, line 61: according to the card, sample type of 2 should resolve to localRecurrence

cvr/src/test/java/org/cbioportal/cmo/pipelines/cvr/CVRClinicalRecordTest.java
```diff
@@ -0,0 +1,70 @@
...
+        String defaultSampleType = record.resolveSampleType(128);
```
sheridancbio, line 62: according to the card, sample type 127 should resolve to the unknown sample type. Other codes should maybe resolve to null / empty / or cause an exception?  I'm not sure what is appropriate.

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/model/staging/CVRClinicalRecord.java
```diff
@@ -370,11 +370,22 @@ public void setMSK_SLIDE_ID(String mskSlideID) {
         this.mskSlideID = mskSlideID;
     }
 
-    private String resolveSampleType(Integer isMetastasis) {
-        if (isMetastasis != null)
-            return isMetastasis == 0 ? "Primary" : "Metastasis";
+    public String resolveSampleType(Integer isMetastasis) {
```
sheridancbio, line 373: Sorry .. one more request ... the parameter to this function should probably be renamed to something like sampleType

ao508: note: we decided to keep the field as is to match the CVR JSON field that we get the value from ("is_metastasis")

sheridancbio: after discussing, I understand that the DMP model still calls this field "isMetastasis", so the nomenclature carries through to this function call. Unless the DMP metadata model is updated (which maybe should be done) we can wait on making this change.

PR #768, opened 2020-07-28
Update Crontab and Fix Auth Header

averyniceday:


import-scripts/precache-portals.py
```diff
@@ -63,8 +63,8 @@ def get_cbio_url(deployment):
     return "http://www.cbioportal.org" if deployment == "cbioportal-backend-master" else "http://www.%s.cbioportal.org" % (DEPLOYMENT_TO_URL_MAP[deployment])
 
 # get DAT for submitting requests for private portal
-def get_authorization_header(deployment):
-    if deployment not in PROTECTED_PORTALS:
+def get_authorization_header(deployments):
+    if all([deployment not in PROTECTED_PORTAL for deployment in deployments]):
```
sheridancbio, line 67: note to self : if there is no protected portal in the list of deployments, we won't need to use the data access token

PR #786, opened 2020-09-17
Detects Non-ASCII characters

divyamadala30:
Detects Unicode characters and provides suggested replacements. Compatible with Python 3. Program has dependency on Unidecode, use "pip install unidecode".

### DIFF
--- a/import-scripts/generate_cdd_attribute_tsv.py
+++ b/import-scripts/generate_cdd_attribute_tsv.py
@@ -63,8 +63,8 @@ import json
 import re
 import requests
 import sys
-#this program requires "pip install unidecode"
 from unidecode import unidecode
+#this program requires "pip install unidecode"
 
 GITHUB_URI_MAPPINGS_FILE_URL = "https://api.github.com/repos/cBioPortal/clinical-data-dictionary/contents/docs/resource_uri_to_clinical_attribute_mapping.txt"
 HEADERS = {"Accept": "application/vnd.github.v4.raw"}
@@ -98,7 +98,16 @@ def create_uri_dictionary(uri_mapping_file):
             data = uri_mapping.split("\t")
             uri_dictionary[data[0].rstrip()] = data[1].rstrip()
     return uri_dictionary
-  
+
+# Looks for non-ASCII characters
+# Uses 'isascii()' attribute for Java 3.7 and over
+def is_ASCII(s):
+    if(hasattr(s, 'isascii')): return s.isascii()
+    for c in s:
+        if (ord(c) > 128): return False
+    return True
+
+    
 # Given a tab delimited file - generate a list of dictionaries where
 # each dictionary is an attribute {NORMALIZED_COLUMN_HEADER: normalized_column_header, DESCRIPTION: description, ...}
 # also adds URI under CONCEPT_ID key
@@ -111,7 +120,7 @@ def load_new_cdd_attributes(new_attributes_file, uri_dictionary):
         sys.exit(2) 
     new_cdd_attributes = []
     invalid_attributes = []
-    non_ascii_values = []
+    non_ascii_characters = []
     last_known_uri_number = get_last_known_uri_number(uri_dictionary)
     with open(new_attributes_file, "r") as f:
         # skip first line
@@ -125,22 +134,21 @@ def load_new_cdd_attributes(new_attributes_file, uri_dictionary):
                 invalid_attributes.append(data[normalized_column_header_position])
             for position in range(len(header)):
                 try:
-                    if (data[position].isascii()):
+                    if (is_ASCII(data[position])):
                         new_attribute[header[position]] = insert_value(header[position], data[position], data[normalized_column_header_position])
                     else:
-                        non_ascii_values.append([data[4],data[position], unidecode(data[position])])
+                        non_ascii_characters.append([data[4], data[position], unidecode(data[position])])
                 except:
                     new_attribute[header[position]] = insert_value(header[position], "", data[normalized_column_header_position])
             new_attribute[CONCEPT_ID_KEY] = get_next_uri(last_known_uri_number)
             new_cdd_attributes.append(new_attribute)
             last_known_uri_number += 1
-            
-     if len(invalid_attributes) > 0:
+    if len(invalid_attributes) > 0:
         print (sys.stderr, "Invalid attributes (already exists/invalid name) in added: " + "\t".join(invalid_attributes) + "... aborting")
         sys.exit(2)
-    if len(non_ascii_values) > 0:
-        print (sys.stderr, "Non-ASCII characters present: ")
-        for i in non_ascii_values: print (i[0] + "," + i[1] + ":" + i[2])
+    if len(non_ascii_characters) > 0:
+        print (sys.stderr, "Non-ASCII characters found in the following: ")
+        for i in non_ascii_characters: print(i[0] + ", " + i[1] + ": " + i[2])
         sys.exit(2)
     return new_cdd_attributes 
 
@@ -164,7 +172,7 @@ def insert_value(key, value, normalized_column_header_name):
         elif key == PRIORITY_KEY:
             return "1"
         else: 
-            print ( sys.stderr, "No name is defined in normalized column header column... aborting")
+            print (sys.stderr, "No name is defined in normalized column header column... aborting")
             sys.exit(2)
     else:
         return value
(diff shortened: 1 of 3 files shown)


import-scripts/generate_cdd_attribute_tsv.py
```diff
@@ -81,7 +83,9 @@
 
 # Extracts credentials for later use
 def get_github_credentials(credentials_file):
-    credentials_line = open(credentials_file,"rb").read().rstrip()
+    print("start func")
```
sheridancbio, line 86: looks like some debugging / status lines were left in ... these should be removed

import-scripts/generate_cdd_attribute_tsv.py
```diff
@@ -97,6 +101,8 @@ def create_uri_dictionary(uri_mapping_file):
             uri_dictionary[data[0].rstrip()] = data[1].rstrip()
     return uri_dictionary
 
+
```
sheridancbio, line 104: Don't need these 2 added lines

import-scripts/generate_cdd_attribute_tsv.py
```diff
@@ -122,14 +129,23 @@ def load_new_cdd_attributes(new_attributes_file, uri_dictionary):
                 invalid_attributes.append(data[normalized_column_header_position])
             for position in range(len(header)):
                 try:
-                    new_attribute[header[position]] = insert_value(header[position], data[position], data[normalized_column_header_position])
+                    if (data[position].isascii()):
+                        new_attribute[header[position]] = insert_value(header[position], data[position], data[normalized_column_header_position])
+                    else:
+                        non_ascii_values.append([data[position], unidecode(data[position])])
```
sheridancbio, line 135: This will help us print out all detected non-ascii characters, but it will not keep track of which Clinical Attribute was involved. This is ok -- but we might also consider tracking the list of affected Clinical Attributes and reporting those .. so that the curators will be able to more quickly find the problem lines.

import-scripts/generate_cdd_attribute_tsv.py
```diff
@@ -122,14 +129,23 @@ def load_new_cdd_attributes(new_attributes_file, uri_dictionary):
                 invalid_attributes.append(data[normalized_column_header_position])
             for position in range(len(header)):
                 try:
-                    new_attribute[header[position]] = insert_value(header[position], data[position], data[normalized_column_header_position])
+                    if (data[position].isascii()):
```
sheridancbio, line 132: this is python3.7 .. which means we would have to upgrade our servers

sheridancbio (approved):
Looks good. Some suggestions / whitespace tweaks requested

import-scripts/generate_cdd_attribute_tsv.py
```diff
@@ -96,7 +98,7 @@ def create_uri_dictionary(uri_mapping_file):
             data = uri_mapping.split("\t")
             uri_dictionary[data[0].rstrip()] = data[1].rstrip()
     return uri_dictionary
-
+
```
sheridancbio, line 101: there are two spaces on this empty line which do not need to be here

import-scripts/generate_cdd_attribute_tsv.py
```diff
@@ -122,14 +125,22 @@ def load_new_cdd_attributes(new_attributes_file, uri_dictionary):
...
+                        non_ascii_values.append([data[4],data[position], unidecode(data[position])])
                 except:
                     new_attribute[header[position]] = insert_value(header[position], "", data[normalized_column_header_position])
             new_attribute[CONCEPT_ID_KEY] = get_next_uri(last_known_uri_number)
             new_cdd_attributes.append(new_attribute)
             last_known_uri_number += 1
-    if len(invalid_attributes) > 0:
-        print >> sys.stderr, "Invalid attributes (already exists/invalid name) in added: " + "\t".join(invalid_attributes) + "... aborting"
+            
+     if len(invalid_attributes) > 0:
```
sheridancbio, line 138: This line is indented 5 spaces instead of 4.

import-scripts/generate_cdd_attribute_tsv.py
```diff
@@ -207,11 +218,12 @@ def update_github_uri_mappings(uri_dictionary, output_directory, github_file_url
     session.auth = (username, password)
     response = session.put(github_file_url, data = json.dumps(data))
     if response.status_code != 200:
-        print >> sys.stderr, "Error encountered when pushing new mappings into github, returned status code: " + str(response.status_code) + "... aborting"
-        sys.exit(2)
-    print "Finished updating uri mapping files in github"
+        print ( sys.stderr, "Error encountered when pushing new mappings into github, returned status code: " + str(response.status_code) + "... aborting")
+        sys.exit(2)
```
sheridancbio, line 222: this line has two extra spaces at the end of the line

sheridancbio (approved):
Just a few extra spaces here and there to remove

import-scripts/generate_cdd_attribute_tsv.py
```diff
@@ -97,6 +99,15 @@ def create_uri_dictionary(uri_mapping_file):
             uri_dictionary[data[0].rstrip()] = data[1].rstrip()
     return uri_dictionary
 
+# Looks for non-ASCII characters
+# Uses 'isascii()' attribute for Java 3.7 and over
+def is_ASCII(s):
+    if(hasattr(s, 'isascii')): return s.isascii()
```
sheridancbio, line 105: please add a space after "if" .. and put the return clause on the following line (indented)

import-scripts/generate_cdd_attribute_tsv.py
```diff
@@ -97,6 +99,15 @@ def create_uri_dictionary(uri_mapping_file):
...
+    for c in s:
+        if (ord(c) > 128): return False
```
sheridancbio, line 107: also use a second line for the return statement here

import-scripts/generate_cdd_attribute_tsv.py
```diff
@@ -122,14 +134,21 @@ def load_new_cdd_attributes(new_attributes_file, uri_dictionary):
...
             new_attribute[CONCEPT_ID_KEY] = get_next_uri(last_known_uri_number)
             new_cdd_attributes.append(new_attribute)
             last_known_uri_number += 1
     if len(invalid_attributes) > 0:
-        print >> sys.stderr, "Invalid attributes (already exists/invalid name) in added: " + "\t".join(invalid_attributes) + "... aborting"
+        print (sys.stderr, "Invalid attributes (already exists/invalid name) in added: " + "\t".join(invalid_attributes) + "... aborting")
+        sys.exit(2)
+    if len(non_ascii_characters) > 0:
+        print (sys.stderr, "Non-ASCII characters found in the following: ")
+        for i in non_ascii_characters: print(i[0] + ", " + i[1] + ": " + i[2])
```
sheridancbio, line 151: Break up this loop into two lines .. where the print statement is indented

sheridancbio (approved):
looks good .. I just recommend breaking up the "one liner" contractions into the standard indented format (for conditionals and loops)

import-scripts/generate_cdd_attribute_tsv.py
```diff
@@ -63,6 +62,8 @@
 import re
 import requests
 import sys
+from unidecode import unidecode
+#this program requires "pip install unidecode"
```
ao508, line 66: let's move this comment to the commit message

ao508 (changes requested):

import-scripts/generate_cdd_attribute_tsv.py
ao508: also just fyi I have now installed this module for our python3 installation for the cbioportalimporter user

import-scripts/generate_cdd_attribute_tsv.py
```diff
@@ -153,7 +172,7 @@ def insert_value(key, value, normalized_column_header_name):
         elif key == PRIORITY_KEY:
             return "1"
         else: 
-            print >> sys.stderr, "No name is defined in normalized column header column... aborting"
+            print (sys.stderr, "No name is defined in normalized column header column... aborting")
```
ao508, line 175: looks like we missed one here :)

ao508, line 175: looks like we missed some of the whitespace adjustments for the print statements in this script

averyniceday:
current unit tests do not pass
#1. No encoding method declared inside the test file (think this is needed because we have special chars to be tested)
#2. Unit tests are now dependent on having a module (unidecode) installed - which we could install temporarily now but think it's best to have a solution that either auto-installs or does not depend on it. 

@sheridancbio

ao508:
@averyniceday on your second point - I installed the module myself for the cbioportal importer. I agree though that we should set up a `requirements.txt` file like other python projects. I think this can be a separate PR though

ao508:
Following up on the failing tests though - I suspect it's because python2 is being used and we need python3 for the ascii checking

import-scripts/test_generate_cdd_attribute_tsv.py
```diff
@@ -0,0 +1,16 @@
+import unittest
```
ao508, line 1: add this to the top of the file to fix the encoding error message
```
# -*- coding: utf-8 -*-
```

sheridancbio:
All requested changes by @ao508 are incorporated ... squashing and merging PR

PR #800, opened 2020-10-14
Disable RAINDANCE fetching and imports

ao508:

### DIFF
--- a/import-scripts/dmp-import-vars-functions.sh
+++ b/import-scripts/dmp-import-vars-functions.sh
@@ -281,30 +281,6 @@ function import_access_ddp_to_redcap {
     return $return_value
 }
 
-# Function for importing raindance cvr files to redcap
-function import_raindance_cvr_to_redcap {
-    return_value=0
-    if ! import_project_to_redcap $MSK_RAINDANCE_DATA_HOME/data_clinical_mskraindance_data_clinical.txt mskraindance_data_clinical ; then return_value=1 ; fi
-    return $return_value
-}
-
-# Function for importing raindance supp date files to redcap
-function import_raindance_supp_date_to_redcap {
-    return_value=0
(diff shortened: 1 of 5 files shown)


import-scripts/dmp-import-vars-functions.sh
```diff
@@ -281,30 +281,6 @@ function import_access_ddp_to_redcap {
     return $return_value
 }
 
-# Function for importing raindance cvr files to redcap
-function import_raindance_cvr_to_redcap {
```
sheridancbio, line 285: we probably can leave these in place .. in case we want to use them again (unless we are sure we won't be importing to raindance again)

import-scripts/fetch-dmp-data-for-import.sh
```diff
@@ -1467,13 +1348,13 @@ if [ $UNLINKED_ARCHER_SUBSET_FAIL -gt 0 ] ; then
     echo -e "$EMAIL_BODY" |  mail -s "UNLINKED_ARCHER Subset Failure: Study will not be updated." $PIPELINES_EMAIL_LIST
 fi
 
-EMAIL_BODY="Failed to merge MSK-IMPACT, HEMEPACT, ARCHER, and RAINDANCE data. Merged study will not be updated."
+EMAIL_BODY="Failed to merge MSK-IMPACT, HEMEPACT, ARCHER, RAINDANCE, and ACCESS data. Merged study will not be updated."
```
sheridancbio, line 1351: :+1:

sheridancbio (approved):
do we know we will not be doing any more raindance updates?

import-scripts/dmp-import-vars-functions.sh
ao508: @sheridancbio this time it's definitely not going to be enabled again - we've finalized the set of raindance ids and confirmed our set matches the DMP-provided set of samples so nothing is missing anymore 🎉

PR #811, opened 2020-11-19
Fix Failing Jenkins Test

averyniceday:
Because of PDX Unit tests which runs through whole pipeline
Adds metadata headers (something was updated in TopBraid)


sheridancbio (approved):
Nice find. Sorry for not realizing that fixing this typo would break our integration tests!  :)

PR #814, opened 2021-01-06
disable debug port use during import

sheridancbio:
debug port use will normally be disabled
debug can be enabled by setting ENABLE_DEBUGGING in import scripts


mandawilson (approved):
Excellent!

PR #845, opened 2021-05-24
Equivalent scripts for EKS Env

averyniceday:
Temporarily renamed the production scripts to `pipelines.*` and made changes to the production scripts so that it's easier to compare what's changed inside GitHub


sheridancbio (commented):
Nothing seemed wrong to me. It is not possible to see any changes to fetch/import-dmp scripts in isolation, so I did not read through the entirety of those.
A couple fine points probably need follow up discussion, including what the plans are for redcap services and for cdd/oncotree. I imagine we will be migrating those into AWS too sometime - but we certainly can leave them running on the pipelines server for now.

sheridancbio:
This is still work-in-progress, so this PR is being left open. We should clean this up after we have a working AWS eks cluster and have checked in the import script changes for real (perhaps through this PR)

PR #854, opened 2021-06-17
chain the cmo data import into the import wrapper

sheridancbio:
- this changes the import timing and order and prevents the cmo imports from overlapping the dmp imports
- now cmo data imports are entirely folded in to the dmp import wrapper
- new schedule:
   - 8pm wrapper starts with dmp fetch (3 hour runtime expected)
   - ~11pm dmp data imports occur (2 hour runtime expected)
   - ~1am cmo data imports occur (runtime varies greatly)
   - next, the pdx data is imported
   - next, the msk-mind data is imported
   - next, the msk-spectrum data is imported
- note: we have left in the protection logic which skips the dmp data import if it is not ready to go by 7am

The first attempt logic (which started the dmp fetch earlier, and then had the wrapper do first the cmo import, and then wait for the dmp fetch to finish before going on to the dmp import) is maintained as a changeset in this branch.


averyniceday (changes requested):
Looks good to me but I wonder if we should cut our losses earlier. The script takes 11:30pm as the last acceptable import time but we've been operating under the assumption that they will end by around 10/11am. 

It's possible it will interfere with other imports we run through the day (or if it starts at 11pm, bleed into the next day). 

Additionally I think we also wanted to avoid restarting the portal during regular hours.

averyniceday (approved):

sheridancbio:
talking with @averyniceday just now, we came to a belief that we don't want the dmp imports happening too late in the day either. Our rule of thumb was that if the dmp imports were not expected to complete by 9am then we would skip over the import attempts. From the little schedule above, if the imports start by 7am, we would expect the dmp study imports to finish by 8:40 ... so we are going to put in logic which says "if we are not ready to go by 7am, then skip dmp imports for today". We have moved up the dmp fetching by an hour and a half (they will start at 00:30 rather than 02:00) and we are going to require that the cmo import script completes before the dmp imports starts. The good thing is that the dmp imports can start as soon as the cmo imports are done and the dmp fetches are done ... so fast fetches and a light cmo import should mean an earlier start than normal.

averyniceday (approved):

sheridancbio:
Latest changes are to swap the order of the dmp data import and the cmo data import. Also, we advance the start time for dmp fetches so that they start at 8pm. We expect these will end around 11pm and then next step in the wrapper script will be to run the dmp data imports. We expect those to last around 2 hours, so around 1am we expect the next step in the wrapper script to run ... which is the cmo data imports. Those will take however long they take (varies greatly) but afterwards the pdx, msk-mind, and spectrum imports happen. If the whole process is not complete by 8pm then the following day's import run is skipped.

PR #855, opened 2021-06-23
restart redis servers when restarting portals

sheridancbio:
update the restart-portal-pods.sh script so that it also will restart appropriate redis services after import to the portal.

optional ```--preserve-cache``` argument prevents cache wiping


inodb (approved):
Thanks 👍

averyniceday (approved):

PR #856, opened 2021-07-19
check for new filenames before merge/subset

sheridancbio:
python logic written in separate library
- looks for lines in datatype sheet which duplicate a previous datatype (new)
- ignore normal sample files (which are ignored by importer)
perform check from subset-impact-data.sh and merge.py and exit on noticing new filename patterns.


import-scripts/subset-impact-data.sh
```diff
@@ -55,6 +55,26 @@ if [ -z $PORTAL_SCRIPTS_DIRECTORY ]; then
 fi
 echo -e "\tPORTAL_SCRIPTS_DIRECTORY="$PORTAL_SCRIPTS_DIRECTORY
 
+# check for presence of new filename pattens (not yet supported) and fail if present
+
+PYTHON_FILENAME_CHECK_CALL="import sys
```
averyniceday, line 60: switch to our usual convention of calling python scripts (see fetch-dmp-data)

import-scripts/updated_filename_for_datatype_test.py
```diff
@@ -0,0 +1,534 @@
...
+    if len(datatype_lines) != len(staging_file_lines):
+        print("Error : datatype_column_content linecount (%d) does not match staging_filename_column_content linecount (%d)" % (len(datatype_lines), len(staging_file_lines)))
+        sys.exit(1);
+    if len(datatype_lines) != len(meta_file_lines):
+        print("Error : datatype_column_content linecount (%d) does not match meta_filename_column_content linecount (%d)" % (len(datatype_lines), len(meta_file_lines)))
+        sys.exit(1);
+    filename2datatype = {}
+    datatype_seen = set([])
+    filename_is_first_listed = set([])
+    for index in range(1,len(datatype_lines)):
```
averyniceday, line 492: if we're looping through items in the list:
`for x in list:`

import-scripts/updated_filename_for_datatype_test.py
```diff
@@ -0,0 +1,534 @@
...
+        datatype = datatype_lines[index].strip();
+        if len(datatype) == 0:
+            continue
+        staging_filename = staging_file_lines[index].strip();
```
averyniceday, line 496: think it's risky to have order-dependent if we're asking other people to maintain this script and update it

averyniceday: or at least have the data pulled in easier rather than manually adding entires

import-scripts/updated_filename_for_datatype_test.py
```diff
@@ -0,0 +1,534 @@
...
+    meta_file_lines = meta_filename_column_content().splitlines()
+    if len(datatype_lines) != len(staging_file_lines):
+        print("Error : datatype_column_content linecount (%d) does not match staging_filename_column_content linecount (%d)" % (len(datatype_lines), len(staging_file_lines)))
+        sys.exit(1);
+    if len(datatype_lines) != len(meta_file_lines):
+        print("Error : datatype_column_content linecount (%d) does not match meta_filename_column_content linecount (%d)" % (len(datatype_lines), len(meta_file_lines)))
+        sys.exit(1);
+    filename2datatype = {}
+    datatype_seen = set([])
+    filename_is_first_listed = set([])
```
averyniceday, line 491: rename this?

import-scripts/updated_filename_for_datatype_test.py
```diff
@@ -0,0 +1,534 @@
...
+meta_drug_treatment_ic50.txt
+meta_drug_treatment_zscore.txt
+meta_mutations_uniprot_canonical_transcripts.txt
+meta_mutations_uniprot_canonical_transcripts.txt
+meta_mutations_uniprot_canonical_transcripts.txt
+meta_gene_matrix.txt
+meta_gene_panel_matrix.txt
+"""
+
+def filename_maps():
```
averyniceday, line 479: rename this

averyniceday, line 479: add comments before each function explaining what data structures are being used and what data they're storing

import-scripts/updated_filename_for_datatype_test.py
```diff
@@ -0,0 +1,534 @@
...
+        if not os.path.isfile(filepath):
+            # ignore subdirectories
+            continue
+        if not file in filename2datatype:
+            # filename/pattern was not found
+            continue
+        if is_normal_sample_filename(file):
+            # normal sample files are skipped in importer
+            continue
+        if not file in filename_is_first_listed:
```
averyniceday, line 531: Think it's dangerous to have order in the google spreadsheet indicate "new vs old"

averyniceday (changes requested):
Just hardcode stuff LOL because of reasons we've talked about!

import-scripts/updated_filename_for_datatype_test.py
sheridancbio: agreed ... the familiar old file names will be hardcoded instead.

import-scripts/subset-impact-data.sh
sheridancbio: done

import-scripts/updated_filename_for_datatype_test.py
```diff
@@ -0,0 +1,534 @@
...
+meta_drug_treatment_zscore.txt
+meta_mutations_uniprot_canonical_transcripts.txt
+meta_mutations_uniprot_canonical_transcripts.txt
+meta_mutations_uniprot_canonical_transcripts.txt
+meta_gene_matrix.txt
+meta_gene_panel_matrix.txt
+"""
+
+def filename_maps():
+    datatype_lines = datatype_column_content().splitlines()
```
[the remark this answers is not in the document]
sheridancbio: the full datatypes sheet content (tab separated values) is now embedded in the script, and the columns are parsed through tab separation.

import-scripts/updated_filename_for_datatype_test.py
sheridancbio: looping on index number removed

import-scripts/updated_filename_for_datatype_test.py
sheridancbio: order dependence is removed ... now there is a separate list of "old filenames" which are safe and unchanging.
The data for the datatypes sheet can now be exported from google sheets to a TSV file and then embedded in the script.

import-scripts/updated_filename_for_datatype_test.py
sheridancbio: this is no longer used. The variable with the "safe" old filenames is : original_datatype_filenames ... which is a simple set of strings.

import-scripts/updated_filename_for_datatype_test.py
sheridancbio: this has been dropped / transformed into a function filenames_in_use_before_update() which returns a simple string set

sheridancbio: done

averyniceday (commented):
add a check for seg files

sheridancbio:
I've added proper checking of wildcard filename patterns .. such as : data_mutations_uniprot_canonical_transcripts_*.txt (a newly added filename pattern)

import-scripts/subset-impact-data.sh
```diff
@@ -55,6 +55,23 @@ if [ -z $PORTAL_SCRIPTS_DIRECTORY ]; then
...
 
+# check for presence of new filename pattens (not yet supported) and fail if present
+
+$PYTHON_BINARY $PORTAL_SCRIPTS_DIRECTORY/study_directory_uses_updated_filenames.py $INPUT_DIRECTORY $STUDY_ID
+if [ $? -eq 0 ] ; then
+    echo "unsupported filenames present - skipping subset" >&2
+    exit 1
+fi
+
+PYTHON_FILENAME_CHECK_CALL="import sys
```
averyniceday, line 66: is this duplicate of the section above? if not, we should have the call be similar to the one before, or if it is remove this

import-scripts/study_directory_uses_updated_filenames.py
```diff
@@ -0,0 +1,538 @@
+#!/usr/bin/env python3
+
+import argparse
+import os
+import sys
+
+def CANCER_STUDY_TAG():
```
averyniceday, line 7: could we just define this as a string?

averyniceday: actually similar comment for all the constants

import-scripts/study_directory_uses_updated_filenames.py
```diff
@@ -0,0 +1,538 @@
...
+    for line in datatype_sheet_lines:
+        if not header_skipped:
+            if line.startswith("datatype"):
+                header_skipped = True
+            continue
+        fields = line.split("\t")
+        datatype = fields[0].strip()
+        staging_filename = fields[5].replace(CANCER_STUDY_TAG(), cancer_study_identifier).strip()
+        meta_filename = fields[9].replace(CANCER_STUDY_TAG(), cancer_study_identifier).strip()
+        if "*" in staging_filename:
```
averyniceday, line 400: could we add a couple examples of these into the comment above? e.g (blah_* would become blah)

import-scripts/study_directory_uses_updated_filenames.py
```diff
@@ -0,0 +1,538 @@
...
+    arguments:
+    path -- the (absolute or relative) filename path to the directory which should be examined
+    cancer_study_identifier -- cancer study stable id [optional]. If not provided, the value is searched for in meta files within <path>
+    """
+    new_filenames_present = False
+    if os.path.isdir(path):
+        path = path + os.path.sep
+    basedir = os.path.dirname(path)
+    filenames = os.listdir(path)
+    if cancer_study_identifier is None:
```
averyniceday, line 496: I think pythonic equivalent would be 
`if not cancer_study_idenfifier`

import-scripts/study_directory_uses_updated_filenames.py
```diff
@@ -0,0 +1,538 @@
...
+    original_datatype_filenames, original_wildcard_patterns = filenames_in_use_before_update(cancer_study_identifier)
+    current_datatype_filenames, current_wildcard_patterns = parse_current_datatypes(cancer_study_identifier)
+    split_original_wildcard_patterns = split_wildcard_patterns(original_wildcard_patterns)
+    split_current_wildcard_patterns = split_wildcard_patterns(current_wildcard_patterns)
+    for filename in filenames:
+        filepath = os.path.join(basedir, filename)
+        if not os.path.isfile(filepath):
+            # ignore subdirectories
+            continue
+        if filename in original_datatype_filenames or matches_any_wildcard_pattern(filename, split_original_wildcard_patterns):
```
averyniceday, line 510: could this and the check below be put in the same line?

averyniceday: or I guess since they are checking for separate specific things it's good to be separate nvm

import-scripts/study_directory_uses_updated_filenames.py
```diff
@@ -0,0 +1,538 @@
...
+    returns True if filename starts with the prefix from, and ends with the suffix from, the split_wildcard_pattern dictionary.
+    """
+    return filename.startswith(split_wildcard_pattern["prefix"]) and filename.endswith(split_wildcard_pattern["suffix"])
+
+def matches_any_wildcard_pattern(filename, split_wildcard_patterns):
+    """
+    iterates through all pattern dictionaries, looking for any match for filename.
+    """
+    for split_wildcard_pattern in split_wildcard_patterns:
+        if (matches_wildcard_pattern(filename, split_wildcard_pattern)):
```
averyniceday, line 474: don't need the parens around the if

import-scripts/study_directory_uses_updated_filenames.py
```diff
@@ -0,0 +1,538 @@
...
+        "meta_protein_quantification_Zscores.txt",
+        "meta_protein_quantification_zscores.txt",
+        "meta_resource_definition.txt",
+        "meta_resource_sample.txt",
+        "meta_resource_study.txt",
+        "meta_rppa.txt",
+        "meta_rppa_Zscores.txt",
+        "meta_timeline.txt"
+        ]
+    filenames_in_use_before_update_set = set()
```
averyniceday, line 374: maybe keep constants as a list of strings then use the replace later in a separate function in its own routine

import-scripts/study_directory_uses_updated_filenames.py
```diff
@@ -0,0 +1,538 @@
...
+def determine_cancer_study_identifier_from_meta_file(filepath):
+    """
+    search file at <filepath> for a cancer_study_identifier and return it if found
+    """
+    CANCER_STUDY_IDENTIFIER_PREFIX = "cancer_study_identifier:"
+    file = open(filepath, 'r')
+    for line in file.readlines():
+        line_without_whitespace = "".join(line.split()).strip()
+        if line.startswith(CANCER_STUDY_IDENTIFIER_PREFIX):
+            cancer_study_identifier = line[len(CANCER_STUDY_IDENTIFIER_PREFIX):].strip()
```
averyniceday, line 429: maybe we shoud just split the line on colons? and take the second part? e.g line.split(':')[1]

averyniceday: something like 
```
cancer_study_identifier = line.split(":")[1].strip() (there should be a strip for both front and back)
if cancer_study_identifier:
blahblahblahblabha
```

import-scripts/study_directory_uses_updated_filenames.py
```diff
@@ -0,0 +1,538 @@
...
+    """
+    search <filenames> in <basedir> for a cancer_study_identifier in any meta filename and return it if found
+    """
+    cancer_study_identifier = None
+    for filename in filenames:
+        if filename.startswith("meta_"):
+            filepath = os.path.join(basedir, filename)
+            if os.path.isfile(filepath):
+                cancer_study_identifier = determine_cancer_study_identifier_from_meta_file(filepath)
+        if cancer_study_identifier is not None:
```
averyniceday, line 444: maybe break the loop once it's set and return cacner_study_identifier at the end of the function

import-scripts/study_directory_uses_updated_filenames.py
```diff
@@ -0,0 +1,538 @@
...
+        if len(pattern_segments) != 2:
+            sys.stderr.write("Error : wildcard filename pattern '" + wildcard_pattern + "' does not have exactly one asterisk\n")
+            sys.exit(1)
+        split_wildcard_pattern = {"prefix" : pattern_segments[0], "suffix" : pattern_segments[1]}
+        split_wildcard_pattern_list.append(split_wildcard_pattern)
+    return split_wildcard_pattern_list
+
+def matches_wildcard_pattern(filename, split_wildcard_pattern):
+    """
+    returns True if filename starts with the prefix from, and ends with the suffix from, the split_wildcard_pattern dictionary.
```
averyniceday, line 465: can we get some examples?

import-scripts/study_directory_uses_updated_filenames.py
averyniceday: additional not necessary but just fun python lol: 

if we want to return true if any of the checks return true we could use the any keyword 

something like return any([matches_wild_card(blahblahblabh) for split_wildcard_pattern in split_wildcard_patterns)]

import-scripts/subset-impact-data.sh
sheridancbio: I thought I did remove this .... yes - no longer needed

import-scripts/study_directory_uses_updated_filenames.py
sheridancbio: 👍

import-scripts/study_directory_uses_updated_filenames.py
sheridancbio: done

import-scripts/study_directory_uses_updated_filenames.py
sheridancbio: done

import-scripts/study_directory_uses_updated_filenames.py
sheridancbio: I've added the newly introduced pattern data_mutations_uniprot_canonical_transcripts_*.txt as an example

import-scripts/study_directory_uses_updated_filenames.py
sheridancbio: Thinking about the possibility of multiple colons on the line (some colons occurring in the value), we would have to join together all parts of the list after the first element, such as
":".join(split_line_fields[1:])

So we chose to recode this by splitting at the position of the first occurrence of ":"

sheridancbio: After looking again, the code already discards all whitespace on the line (from the meta file). It does this before tests are performed, so the prefix "cancer_study_identifier:" would be seen and could be stripped off of the line based on length, which would yield just the value string (minus any spaces). Since cancer studies should have no spaces, there should be no problem. So I think I'll keep the code as is.

import-scripts/study_directory_uses_updated_filenames.py
sheridancbio: done

import-scripts/study_directory_uses_updated_filenames.py
sheridancbio: the example added for parse_current_datatypes now is added here to show the use of the split wildcard pattern.

import-scripts/study_directory_uses_updated_filenames.py
sheridancbio: 👍

import-scripts/study_directory_uses_updated_filenames.py
sheridancbio: done

import-scripts/study_directory_uses_updated_filenames.py
sheridancbio: yes .. probably clearer to have them separate

sheridancbio:
We decided together that the effort to properly determine the cancer_study_identifer was not feasible in all use cases. The pdx integration tests for example perform a merge step on derived (subsetted) studies which do not contain the needed properties in meta files. Also, the cancer_study_identifier which is passed into the merge.py script was seen to be the identifier for the output study .. not the input study. So it could not be used to validate the input filenames after all for filenames of pattern <CANCER_STUDY>_data_cna_hg19.seg.  So we have punted on detection of "correctness" based on the cancer_study_identifier. The check was still valuable for other filename patterns however, so it is being merged as a check for merge.py and subset_impact_data.sh - without the processing of cancer_study_identifier.

import-scripts/study_directory_uses_updated_filenames.py
```diff
@@ -0,0 +1,511 @@
...
+        new_filenames_present = True
+    return new_filenames_present
+
+def main(args):
+    """
+    scans a study directory for any novel filenames which are not yet supported by the merge scripts
+    returns 0 if novel/updated filenames are preset, 1 otherwise
+    """
+    if study_directory_uses_updated_filenames(args.cancer_study_path):
+        sys.exit(0);
```
averyniceday, line 503: swap these two exit codes

import-scripts/subset-impact-data.sh
```diff
@@ -55,6 +55,14 @@ if [ -z $PORTAL_SCRIPTS_DIRECTORY ]; then
 fi
 echo -e "\tPORTAL_SCRIPTS_DIRECTORY="$PORTAL_SCRIPTS_DIRECTORY
 
+# check for presence of new filename pattens (not yet supported) and fail if present
+
+$PYTHON_BINARY $PORTAL_SCRIPTS_DIRECTORY/study_directory_uses_updated_filenames.py -d $INPUT_DIRECTORY
+if [ $? -eq 0 ] ; then
```
averyniceday, line 61: checking for -eq 0 (which is technically working right now for the main test, but we still have a nonzero exit code for a failed test in one of the other checks) 

to keep it simple, all failed tests (meaning new files present, or anything else) should stick to a non-zero exit code and the check can be for -ne 0

averyniceday (changes requested):
fix exit codes!

PR #857, opened 2021-07-20
add authentication to k8s related scripts

averyniceday:


sheridancbio (commented):
I think authenticate_service_account.sh needs to be checked in ... also, this needs to work both on the pipelines production server and also the eks cluster (restart-portal-pods.sh does). So the authenticate script needs to "exit gracefully" on pipelines.

import-scripts/delete_slow_to_terminate_pods.sh
```diff
@@ -183,6 +183,7 @@ MY_FLOCK_FILEPATH="/data/portal-cron/cron-lock/delete_slow_to_terminate_pods.loc
         mv $temp_file $POD_TERMINATING_TIMESTAMP_FILE
     }
 
+    /data/portal-cron/scripts/authenticate_service_account.sh
```
sheridancbio, line 186: I wonder why we are calling in the scripts subdirectory. The script is commited in portal-configuration repo, so why not /data/portal-cron/git-repos/portal-configuration/... ?

import-scripts/restart-portal-pods.sh
```diff
@@ -8,7 +8,10 @@ if ! which $KUBECTL_BINARY > /dev/null 2>&1 ; then
     exit 1
 fi
 
+$PORTAL_HOME/scripts/authenticate_service_account.sh
```
sheridancbio, line 11: do we need this when we call it below?

sheridancbio (approved):

PR #859, opened 2021-07-29
Separate EKS vs Pipelines Scripts

averyniceday:
Changes to automation environment for differing paths as well as the mycrontab

updated crontab check to run against different mycrontab based on hostname

For deploying, copy the equivalent scripts based on what machine

### DIFF
--- a/import-scripts/monitor-crontab-version.sh
+++ b/import-scripts/monitor-crontab-version.sh
@@ -1,16 +1,7 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-# function for sending notification emails
-send_email_notification() {
-    diff_report=$1
-    ### FAILURE EMAIL ###
-    EMAIL_BODY="The current (pipelines) crontab is out of sync with the crontab in github.\n\n $diff_report"
-    echo -e "Sending email\n$EMAIL_BODY"
-    echo -e "$EMAIL_BODY" | mail -s "Alert: Crontab out of sync on $HOSTNAME" cbioportal-pipelines@cbio.mskcc.org
-}
-
-if [ -z $PORTAL_HOME ] | [ -z $PIPELINES_CONFIG_HOME ] ; then
-    echo "monitor-crontab-version.sh could not be run: missing environment variables must be set using automation-environment.sh"
+if [ -z "$PORTAL_HOME" ] | [ -z "$PIPELINES_CONFIG_HOME" ] ; then
+    echo "monitor-crontab-version.sh could not be run: missing environment variables must be set using automation-environment.sh" >&2
     exit 1
 fi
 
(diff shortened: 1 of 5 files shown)


import-scripts/monitor-crontab-version.sh
```diff
@@ -14,20 +14,30 @@ if [ -z $PORTAL_HOME ] | [ -z $PIPELINES_CONFIG_HOME ] ; then
...
-GITHUB_CRONTAB_URL="https://api.github.com/repos/knowledgesystems/cmo-pipelines/contents/import-scripts/mycrontab"
+GITHUB_PIPELINES_CRONTAB_URL="https://api.github.com/repos/knowledgesystems/cmo-pipelines/contents/import-scripts/pipelines/mycrontab"
+GITHUB_PIPELINES_EKS_CRONTAB_URL="https://api.github.com/repos/knowledgesystems/cmo-pipelines/contents/import-scripts/pipelines_eks/mycrontab"
 GITHUB_AUTHORIZATION=$(cat $PIPELINES_CONFIG_HOME/git/git-credentials | sed 's/https:\/\///; s/@github\.com//')
 CURRENT_CRONTAB_FILE="$PORTAL_HOME/tmp/current_crontab"
 GITHUB_CRONTAB_FILE="$PORTAL_HOME/tmp/git_crontab"
 REQUIRED_USERNAME="cbioportal_importer"
-REQUIRED_HOSTNAME="pipelines.cbioportal.mskcc.org"
+PIPELINES_HOSTNAME="pipelines.cbioportal.mskcc.org"
+PIPELINES_EKS_HOSTNAME="ip-10-1-21-42.mskcc.root.mskcc.org"
```
sheridancbio, line 24: maybe we should create a symbolic hostname / dns record for the eks pipelines node?  Otherwise this code will break if we ever redeploy the eks pipelines node and it is assigned a new IP address.

import-scripts/pipelines/automation-environment.sh
```diff
@@ -0,0 +1,136 @@
+#!/bin/bash
```
sheridancbio, line 1: Note to self : This location is not the deployment location ... the deployer should copy/move the appropriate script into the upper level scripts directory (selecting the correct one based on which platform we are deploying to)

import-scripts/pipelines/mycrontab
```diff
@@ -0,0 +1,42 @@
+MAILTO=cbioportal-pipelines@cbio.mskcc.org
```
sheridancbio, line 1: Note to self: although mycrontab is not a called script, this file perhaps should be also "deployed" to the upper level scripts directory .. to sit beside automation_environment. (as a reference to maintainers)

sheridancbio (approved):
Looks fine for now ... but up until now we have operated using the assumption that the import-scripts subdirectory would be deployed (and exactly match) what is in /data/portal-cron/scripts ... this helps us quickly check if any uncommitted updates were made in the scripts directory, or if any undeployed PRs were merged into master. I hope we can return to that assumption once we return to having a single pipelines crontab running (only in the eks cluster). I also think we should not drift farther away from that assumption (even though we are breaking it for now).

PR #867, opened 2021-08-03
Fix bug with redcap logging

ao508:
Signed-off-by: Angelica Ochoa <ochoaa@mskcc.org>


cvr/pom.xml
```diff
@@ -14,12 +14,6 @@
   </parent>
 
   <dependencies>
-    <dependency>
```
ao508, line 17: this is already in the master pom so I moved it out of here

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/BatchConfiguration.java
```diff
@@ -772,4 +784,76 @@ public JobExecutionDecider svFusionsStepExecutionDecider() {
     public JobExecutionDecider segStepExecutionDecider() {
         return decideStepExecutionByDatatypeForStudyId("seg");
     }
+
+    // general spring batch configuration
+    @Value("org/springframework/batch/core/schema-drop-sqlite.sql")
+    private Resource dropRepositoryTables;
+
+    @Value("org/springframework/batch/core/schema-sqlite.sql")
+    private Resource dataRepositorySchema;
```
ao508, line 793: all of the changes in the BatchConfiguration classes are the same. 

These are required beans that need to be explicitly defined in the newer version of Spring Batch that we're using now

gene/src/main/java/org/mskcc/cmo/ks/GeneDataPipeline.java
```diff
@@ -49,14 +49,14 @@
 
 @SpringBootApplication
 public class GeneDataPipeline {
-    
-    private static final Log LOG = LogFactory.getLog(GeneDataPipeline.class);
-    
+
+    private final static Logger LOG = Logger.getLogger(GeneDataPipeline.class);
```
ao508, line 53: standardizing the logger used

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/BatchConfiguration.java
sheridancbio: ✅

sheridancbio (approved):
these logger libraries continue to haunt us ..
nice work

PR #884, opened 2021-09-03
use the login shell path for automation

sheridancbio:
Scripts invoked via crontab do not receive environment initialization via
/etc/profile (or ~/.bash_profile).
This is because crontab executions are non-interactive non-login invocations.
Therefore any required tools / applications which are not on the very basic
(/usr/bin:/bin) PATH are not available during script execution.

This change sets PATH (of the current shell) to match the PATH of a login
shell (which does get /etc/profile and also ~/.bash_profile initialization).
So long as the system is properly configured so that the user connected to
the crontab has a correct PATH set for their login shell, the pipelines
scripts will function correctly: this removes system-specificity in scripts.


sheridancbio:
This has been tested on the eks node and functions correctly.

PR #886, opened 2021-09-09
clear the persistence cache via api call

sheridancbio:
clear the persistence cache via api call
    
    - drop the tomcat / war restart functions
    - add functions to clear persistence caches
        - uses curl to call the cache clearing endpoint in the portal
        - a file reference holding the cache_api_key is introduced (PORTAL_HOME/pipelines-credentials/cache.api.key)
        - some temporary portal URLs are present because we have not completed rollout yet
        - portal caches are cleared in groups, based (mainly) on database/tomcat {msk, external, triage, public, genie, genie-archave}
        - failure status is reported if any
    - replace calls to restart functions with calls to clear cache functions in all import data scripts
    - drop the old "gdac-aws" import script (early attempt to run imports into the KS AWS account)
    - drop cbioportal-restful-login-counter : this relied on actual tomcat log files (which are no longer available)
    - drop update-gene-data.sh : this is obsolete - we are now manually controlling gene table updates using HGNC databases
    
    Co-authored-by: Avery Wang <18199796+averyniceday@users.noreply.github.com>
    Co-authored-by: Manda Wilson <1458628+mandawilson@users.noreply.github.com>
    Co-authored-by: Robert Sheridan <7747489+sheridancbio@users.noreply.github.com>


mandawilson (approved):

import-scripts/dmp-import-vars-functions.sh
```diff
@@ -90,6 +90,13 @@ function sendImportSuccessMessageMskPipelineLogsSlack {
     curl -X POST --data-urlencode "payload={\"channel\": \"#msk-pipeline-logs\", \"username\": \"cbioportal_importer\", \"text\": \"MSK cBio pipelines import success: $STUDY_ID\", \"icon_emoji\": \":tada:\"}" $SLACK_PIPELINES_MONITOR_URL
 }
 
+# Function for alerting slack channel of clear cache failures
+function sendClearCacheFailureMessageMskPipelineLogsSlack {
+    CACHE_GROUP_NAME=$1
+    SOURCE_SCRIPT_NAME=$2
+    curl -X POST --data-urlencode "payload={\"channel\": \"#msk-pipeline-logs\", \"username\": \"cbioportal_importer\", \"text\": \"Failure to clear the persistence cache for $CACHE_GROUP_NAME portals during import script $SOURCE_SCRIPT_NAME\", \"icon_emoji\": \":plunger:\"}" $SLACK_PIPELINES_MONITOR_URL
```
averyniceday, line 97: unless we're really planning to monitor processes to this extent, I think this might just add to noise in the slack channel.

import-scripts/dmp-import-vars-functions.sh
```diff
@@ -312,82 +319,46 @@ function filter_derived_clinical_data {
...
-        if ! ssh ${SSH_OPTIONS} ${TOMCAT_HOST_USERNAME}@${server} "umask 000 ; touch ${TOMCAT_SERVER_RESTART_PATH}" ; then
-            failed_restart_server_list[${#failed_restart_server_list[*]}]=${server}
-        fi
-    done
-    if [ ${#failed_restart_server_list[*]} -ne 0 ] ; then
-        EMAIL_BODY="Attempt to trigger a restart of the $TOMCAT_SERVER_DISPLAY_NAME server on the following hosts failed: ${failed_restart_server_list[*]}"
-        echo -e "Sending email $EMAIL_BODY"
-        echo -e "$EMAIL_BODY" | mail -s "$TOMCAT_SERVER_PRETTY_DISPLAY_NAME Restart Error : unable to trigger restart" $PIPELINES_EMAIL_LIST
-    fi
+function clearPersistenceCachesForExternalPortals() {
```
averyniceday, line 339: Considering that these portals share the same db and all get refreshed together at once - maybe we just switch the deployments to use the same redis db and clear that single one? 

Would save both on the memory being used and the time restarting each cache

averyniceday (changes requested):
Some comments about whether all the logging is necessary + missing sources for the restart functions in the other import-* scripts (outside of msk and cmo)

import-scripts/dmp-import-vars-functions.sh
sheridancbio: this was changed to a mail message to cbioportal-pipelines

import-scripts/dmp-import-vars-functions.sh
sheridancbio: I like this idea ... the clearCache script has temporary URLs listed anyway and will need a subsequent update/PR. I suggest that we experiment with the deployments and try out this suggestion. When the deployment configuration has stabilized we can adjust the scripts so that the appropriate website URLs are called for clearing the cache. If only one external portal redis is running, we can cut down the number of URLs which are hit.

PR #888, opened 2021-09-15
restart redis services to clear persistence cache

sheridancbio:
- rather than using the web api endpoint to clear persistence cache, we return to restarting the redis persistence pods
- we have also re-introduced the restart of the web application pods (backends)


import-scripts/pipelines_eks/automation-environment.sh
```diff
@@ -11,7 +11,7 @@ export PYTHON3_BINARY=/usr/bin/python3
 export MAVEN_BINARY=/opt/apache-maven-3.8.1/bin/mvn
 export HG_BINARY=/usr/bin/hg
 export GIT_BINARY=/usr/bin/git
-export PATH=/opt:/usr/bin:/data/tools/python2.7/bin:/data/tools/python3.4.10/bin:$PATH
+export PATH=$(bash --login -c 'echo $PATH')
```
averyniceday, line 14: not familiar with this but off google looks like it loads in what's in bash_profile?

import-scripts/clear_cbioportal_persistence_cache.sh
```diff
@@ -23,52 +18,19 @@ portal_to_deployment_map["triage"]="eks-triage"
...
-portal_to_url["prostate"]="https://prostate.cbioportal.aws.mskcc.org"
-portal_to_url["sclc"]="https://sclc.cbioportal.aws.mskcc.org"
-portal_to_url["target"]="https://target.cbioportal.aws.mskcc.org"
+unset portal_to_cache_service_list
+declare -A portal_to_cache_service_list
+portal_to_cache_service_list["public"]="cbioportal-public-persistence-redis-master cbioportal-public-persistence-redis-slave"
+portal_to_cache_service_list["genie-public"]="cbioportal-persistence-redis-genie-master cbioportal-persistence-redis-genie-slave"
+portal_to_cache_service_list["genie-private"]="cbioportal-persistence-redis-genie-master cbioportal-persistence-redis-genie-slave"
+portal_to_cache_service_list["genie-archive"]=""
+portal_to_cache_service_list["triage"]="cbioportal-persistence-redis-master cbioportal-persistence-redis-replicas"
```
averyniceday, line 29: triage will be it's own I believe

averyniceday (changes requested):
Looks good to me, just small change (putting triage on it's own persistence cache) 👍

import-scripts/pipelines_eks/automation-environment.sh
sheridancbio: bash can start up in a few different modes, based on whether it is connected to an interactive terminal (like when we are manually logged in) or not. So on startup, bash is considered one of {interactive, non-interactive} and also one of {login, non-login}. When a bash process is launched by crontab it is considered non-interactive and non-login and the normal bash_profile startup scripts are not executed. The only PATH which is set is a very rudimentary default -- it is only PATH="/usr/bin:/bin". However, you can tell bash to start up as a login shell with the command line option "--login" ... and then the normal bash_profile initialization happens. So this code launches a new bash subshell as a login shell and extracts the initialized PATH environment variable from the subshell. It then sets the top level shell to have the same PATH.

import-scripts/clear_cbioportal_persistence_cache.sh
sheridancbio: Adjusting now. There is now a running triage-cbioportal-persistence-redis helm installation and triage is using it.

averyniceday (approved):

PR #911, opened 2021-11-09
Use current timezone midnight

sheridancbio:
- instances are now midnight in the current (local) timezone. This corrects errors where a date was constructed on a date in daylight savings time (getting a midnight date then), but is to be used for time difference compared to a standard time midnight.


ddp/ddp_pipeline/src/main/java/org/mskcc/cmo/ks/ddp/pipeline/util/DDPUtils.java
```diff
@@ -422,12 +431,27 @@ public static String resolveTimelineEventDateInDays(String referenceDate, String
...
+     *           daylight saving time shifts. So, when constructing a java date instance, the constructed Date may be based on a different timezone (such as when the current timezone is EST, and the constructed
+     *           date was during the time of the year when daylight savings time is in effect. If that is so, the constucted Date object may be midnight in a DST timezone (EDT), and a correction factor may be needed
+     *           to shift the underlying millisecond count since epoc so that the instant corresponds to midnight in EST on that date (even though on the actual date the locale was using EDT).
+     *           This adjustment factor is passed in as a second argument to computeDaysFromDateUsingLocalMidnight(), which will adjust the counting of passed days accordingly.
+     *   TODO: the adjustment factor could be removed if getDateInDays(String dateValue) were rewritten to use SimpleDateFormat("yyyy-MM-dd z"), which would take a time zone specifier. In that case, the parsed
```
mandawilson, line 451: 👍

mandawilson (approved):
Excellent!

PR #917, opened 2021-12-02
fix kill-import script

sheridancbio:
- the kill-import-cmo-data-triage.sh script would kill itself,
    generating a crontab failure email and leading to a stuck
    "import-trigger/triage-import-killing" file in existence,
    blocking import start requests


mandawilson (approved):
:laughing:

PR #922, opened 2021-12-10
Allow Different Clusters for Persistence Clearing

averyniceday:
- k8s config file stored separately
- specific portals will set the k8s config arg 
- commands will use --kubeconfig setting to access other cluster


sheridancbio (approved):
👍 nice and simple

PR #923, opened 2021-12-13
fix for persistence cache clearing on pipelines

sheridancbio:
follow up for #922 to handle pipelines server [which does not define or need special config file]


averyniceday:
did this bug come up? We shouldn't be importing anything from pipelines so this shouldn't be needed

averyniceday (approved):

PR #925, opened 2021-12-16
Script to frequently examine the database and reset persistence cache

sheridancbio:
- new script introduced
- property parsing functionality removed from dmp-import-vars-functions.sh and expanded

See comments in clear-persistence-caches-for-changed-studies.sh script for overview/summary of functionality


import-scripts/clear-persistence-caches-for-changed-studies.sh
```diff
@@ -0,0 +1,450 @@
...
+# usage : search_for_studies_with_new_modifications_in_db db_label
+function search_for_studies_with_new_modifications_in_db() {
+    db="$1"
+    tmp_study_subdir="$TMP_DIR_ROOT/$db"
+    create_directory_if_necessary "$tmp_study_subdir"
+    previous_study_info_from_db_filepath="$TMP_DIR_ROOT/${db}_study_info"
+    new_study_info_from_db_filepath="$TMP_DIR_ROOT/${db}_study_info_new"
+    if ! [ -f "$previous_study_info_from_db_filepath" ] ; then
+        # if we have lost the previous study info, we cannot tell what has been modified. Prime the study info and trigger a full cache reset
+        write_current_study_list_from_db "$db" "$previous_study_info_from_db_filepath"
```
n1zea144, line 240: Check return status here.

import-scripts/clear-persistence-caches-for-changed-studies.sh
```diff
@@ -0,0 +1,450 @@
...
+
+# phase one top level logic - sets up / cleans up / creates the task triggers and status directories/content for every managed database
+# also has logic to override study cache reset with a full cache reset when task backlog gets stale
+function detect_and_maintain_needed_actions_for_databases() {
+    create_directory_if_necessary "$TMP_DIR_ROOT"
+    for portal_db in $PORTAL_DB_LIST ; do
+        clear_completed_triggers "$portal_db"
+        if ! database_is_triggered_as_needing_reset "$portal_db" ; then
+            if any_trigger_has_exceeded_maximum_wait "$portal_db"; then
+                trigger_db_as_needing_reset "$portal_db"
```
n1zea144, line 285: continue to the next loop iteration here.

import-scripts/clear-persistence-caches-for-changed-studies.sh
```diff
@@ -0,0 +1,450 @@
...
+    fi
+    return 1
+}
+
+# clears the persistence cache, either using the cache api for study cache reset tasks, or the kubectl based script for full database cache clearing tasks
+# proper function depends on having the selected cache reset work task set in global environment variables
+# usage : perform_selected_cache_reset_task
+function perform_selected_cache_reset_task() {
+    # if (unlikely) task has been successfully performed while we selected it, exit
+    if trigger_filepath_is_marked_as_succeeded "$selected_cache_reset_work_task_db" "$selected_cache_reset_work_task_trigger_filepath" ; then
```
n1zea144, line 372: Is this really needed?

import-scripts/clear-persistence-caches-for-changed-studies.sh
```diff
@@ -0,0 +1,450 @@
...
+    fi
+    # handle study cache reset if triggered and able
+    if ! database_can_clear_per_study_cache "$selected_cache_reset_work_task_db" ; then
+        return 1 # not able - fail
+    fi
+    # trigger is per-study cache reset (by deduction)
+    study_id="${trigger_filename%.*}"
+    cache_api_key="${portal_db_to_cache_api_key[$selected_cache_reset_work_task_db]}"
+    cache_api_base_uri="${PORTAL_DB_TO_PORTAL_API_BASE_URI[$selected_cache_reset_work_task_db]}"
+    curl -X DELETE -H "X-API-KEY:$cache_api_key" -v -i "$cache_api_baseuri/" "$cache_api_base_uri/cache/$study_id"
```
n1zea144, line 390: Double check why two baseuri's are here.

n1zea144 (changes requested):

sheridancbio:
Added example/explanation to the docs with example trigger/status filename references.

import-scripts/clear-persistence-caches-for-changed-studies.sh
sheridancbio: Done. If the mysql command fails, a log message is generated and this function sees the failure status and returns failure too.

import-scripts/clear-persistence-caches-for-changed-studies.sh
sheridancbio: Done

import-scripts/clear-persistence-caches-for-changed-studies.sh
sheridancbio: Not really. The original thinking was that multiple processes can be running in PHASE TWO and it is hypothetically conceivable that one process might stop at the time between selecting the triggered task to perform and the obtaining of the lock to do the task. However, this case does not need to be covered because:
- after the trigger file is selected, the next step is to lock the file, making an interruption very unlikely
- it is also extremely unlikely that during this momentary vulnerability, another process could acquire the lock, accomplish the entire task, mark successful status and release the lock before this process advanced at all. It is practically certain that the other process would be holding the lock at the time we attempted to get it, so we would fail to obtain the lock and thereby exit.
- even if the envisioned scenario managed to happen, the only effect would be to clear the cache a second time for the trigger, which is not harmful.
Function and function call removed.

import-scripts/clear-persistence-caches-for-changed-studies.sh
sheridancbio: that was an error. Thanks. That was fixed, and I removed the "-v" (verbose) option and the "-i" (include response headers) option to help reduce logging noise. I've also sent the standard output to /dev/null.

sheridancbio: Actually, I tested and curl returns status 0 (success) even when the endpoint returns a failure message. Instead of redirecting to null, I pipe to grep to see if a "success" word is present in output. The word "Flushed" will appear in the output on success, so I test for it.

sheridancbio:
I have also added an entry in mycrontab, and removed the call to the full recache at the end of the public import run script. With this running, that call will be redundant.

n1zea144 (approved):

PR #926, opened 2022-01-03
Log HTML Error Messages

averyniceday:
based on #907


sheridancbio (approved):
These all look good. I thought there were a few more to cover, but I think @averyniceday identified that the other cases (if they fail) would bring the entire fetch process to a halt so the errors would not be buried in a log where we might not notice.

PR #930, opened 2022-02-04
Add support for v2 schema

averyniceday:
Using existing SV model for majority of changes
Add a check for Gene 1 (required not null on CVR end) to determine if v1/v2
Add site1_exon/site2_exon support

### DIFF
--- a/cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/model/CVRSvVariant.java
+++ b/cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/model/CVRSvVariant.java
@@ -63,10 +63,12 @@ import javax.annotation.Generated;
     "paired_end_read_support",
     "site1_chrom",
     "site1_desc",
+    "site1_exon",
     "site1_gene",
     "site1_pos",
     "site2_chrom",
     "site2_desc",
+    "site2_exon",
     "site2_gene",
     "site2_pos",
     "split_read_support",
@@ -110,6 +112,8 @@ public class CVRSvVariant {
     private String site1Chrom;
     @JsonProperty("site1_desc")
     private String site1Desc;
+    @JsonProperty("site1_exon")
+    private String site1Exon;
     @JsonProperty("site1_gene")
     private String site1Gene;
     @JsonProperty("site1_pos")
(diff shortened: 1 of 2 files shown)


cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/model/CVRSvVariant.java
```diff
@@ -159,36 +165,43 @@ public CVRSvVariant() {
...
-    * @param mapq
-    * @param variantStatusName
-    * @param site1Desc
-    * @param comments
-    * @param site2Chrom
-    * @param site2Gene
-    * @param annotation
-    */
-    public CVRSvVariant(String annotation, String breakpointType, String comments, String confidenceClass, String connType, String connectionType, String eventInfo, String mapq, String normalReadCount, String normalVariantCount, String pairedEndReadSupport, String site1Chrom, String site1Desc, String site1Gene, String site1Pos, String site2Chrom, String site2Desc, String site2Gene, String site2Pos, String splitReadSupport, String svClassName, String svDesc, String svLength, String svVariantId, String tumorReadCount, String tumorVariantCount, String variantStatusName) {
+     * @param annotation
```
averyniceday, line 168: this was auto-generated from IntelliJ, I don't think it hurts to update it? don't think it's being used anywhere

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/model/staging/CVRSvRecord.java
```diff
@@ -105,7 +109,61 @@ public CVRSvRecord(CVRSvVariant variant, String sampleId) {
         this.tumor_read_count = variant.getTumor_Read_Count();
         this.tumor_variant_count = variant.getTumor_Variant_Count();
         this.variant_status_name = variant.getVariant_Status_Name();
+
+        // CVR confirmed Site1_GENE/Gene1 is a NOT_NULL field
+        // Use Gene1 to test whether v1 or v2 schema
+        if (variant.getSite1_Gene() != null && !variant.getSite1_Gene().isEmpty()) {
+            this.site1_gene = variant.getSite1_Gene();
```
mandawilson, line 116: Delete this.site1_gene and this.site2_gene initialisation above.

mandawilson: Check for calls to setGene1 and setGene2 to see if site1_gene and site2_gene needs setting in that case.

mandawilson: Avery looked it over and said these are only called once and only for the data_sv.txt not the cvr json file (so this is correct behaviour).

PR #936, opened 2022-02-16
Fix Issue with DDP Vital Status Fields

averyniceday:
INT_DOD and INT_CONTACT were not calculating intervals and were just directly converting dates
Added the following:
- utility function for finding the number of days between two date strings
- anonymizing function for hiding number of days if it is greater than 90 years or less than 18 years

### DIFF
--- a/ddp/ddp_pipeline/src/main/java/org/mskcc/cmo/ks/ddp/pipeline/util/DDPUtils.java
+++ b/ddp/ddp_pipeline/src/main/java/org/mskcc/cmo/ks/ddp/pipeline/util/DDPUtils.java
@@ -166,6 +166,17 @@ public class DDPUtils {
         return anonymizePatientAge(age.intValue());
     }
 
+    /**
+     * Given two date strings, find the number of days that has elapsed between the two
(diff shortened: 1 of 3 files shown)


ddp/ddp_pipeline/src/main/java/org/mskcc/cmo/ks/ddp/pipeline/util/DDPUtils.java
```diff
@@ -166,6 +166,21 @@ public static String resolvePatientCurrentAge(DDPCompositeRecord compositeRecord
         return anonymizePatientAge(age.intValue());
     }
```
ao508, line 168: Maybe add a javadoc here to save our future selves (or future victims) a headache :)

ao508 (commented):
Just one comment regarding adding a javadoc to the new method added. 

One other follow up however: are there existing unit tests that we can update to use this method? Or maybe add?

ao508 (approved):

ao508:
merging once the tests pass!

PR #949, opened 2022-04-19
Revert "Revert "Add NATS messaging to CVR "consume" step ""

sheridancbio:
Restores knowledgesystems/cmo-pipelines#945


sheridancbio:
re-introduced #945 content now that production smile server is accessible from our fetcher machines

PR #950, opened 2022-04-25
Oncokbtype support

averyniceday:


cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/model/staging/CVRSvRecord.java
```diff
@@ -93,12 +93,25 @@ public class CVRSvRecord {
...
+
     public CVRSvRecord() {
     }
 
     public CVRSvRecord(CVRSvVariant variant, String sampleId) {
         this.sampleId = sampleId;
-        this.annotation = variant.getAnnotation();
+        // special handling for joining annotation and oncokb_type into one field in the data SV file
+        // format is [annotation];ONCOKB_TYPE:[oncokb_type] - no semi-colon if either value is null/empty
+        String annotation = null;
```
sheridancbio, line 105: it's confusing to have a local variable called annotation and also a data member called annotation. I suggest calling the local variable "typedAnnotation" or "revisedAnnotation". But I'm not clear on why it is necessary to insert the value of the oncokbType member into the annotation member. It seems to me that it would be more proper for the client to construct the composite string. Otherwise ... why do we even need an oncokbType data member if the value is embedded in annotation? If the answer is "so that we can strip off the type in order to get back the original annotation then I would think twice about this approach.

sheridancbio (commented):
Code seems fine but I don't understand why one field gets embedded inside another.

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/model/staging/CVRSvRecord.java
sheridancbio: Ok, I misunderstood initially. There is no oncokbType field in CVRSvRecord ... rather we are merging together the two fields (oncokbType and annotation) from the variant during construction of the CVRSvRecord object. That seems fine.

I still would rename the local variable for clarity, though.

averyniceday:
checked latest CVR JSON - there is no oncokb_type field

PR #962, opened 2022-06-24
Remove Fusion Support and Replace with SV Support

averyniceday:


sheridancbio (approved):
lgtm ... seems to be a straightforward switchover.

I guess we already had the getStandardSvHeader() function implemented previously in CVRSvRecord, which is nice/better than having local code.

PR #963, opened 2022-07-01
Update get/set and field names to match new SV Format

averyniceday:
Header fields have underscore, need to incorporate into field names and propagate out


sheridancbio (approved):
👍

PR #966, opened 2022-08-04
add utility for standardizing sv variants

sheridancbio:
- gene references (for intragenic cases)
- event info (filled in when blank)

### DIFF
--- a/cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/sv/CVRSvDataReader.java
+++ b/cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/sv/CVRSvDataReader.java
@@ -32,19 +32,20 @@
 
 package org.cbioportal.cmo.pipelines.cvr.sv;
 
-import org.cbioportal.cmo.pipelines.cvr.model.staging.CVRSvRecord;
-import org.cbioportal.cmo.pipelines.cvr.*;
-import org.cbioportal.cmo.pipelines.cvr.model.*;
-
 import java.io.*;
 import java.util.*;
 import org.apache.log4j.Logger;
+import org.cbioportal.cmo.pipelines.cvr.*;
+import org.cbioportal.cmo.pipelines.cvr.sv.SvUtilities;
+import org.cbioportal.cmo.pipelines.cvr.model.*;
+import org.cbioportal.cmo.pipelines.cvr.model.staging.CVRSvRecord;
 import org.springframework.batch.item.*;
 import org.springframework.batch.item.file.FlatFileItemReader;
 import org.springframework.batch.item.file.mapping.DefaultLineMapper;
 import org.springframework.batch.item.file.transform.DelimitedLineTokenizer;
 import org.springframework.beans.factory.annotation.*;
 import org.springframework.core.io.FileSystemResource;
+
 /**
  *
  * @author heinsz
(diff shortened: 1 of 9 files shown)


cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/sv/GMLSvDataReader.java
```diff
@@ -108,7 +109,12 @@ private void processJsonFile() {
             if (samples != null) {
                 for (GMLCnvIntragenicVariant cnv : result.getCnvIntragenicVariantsGml()) {
                     for (String sampleId : samples) {
-                        CVRSvRecord record = new CVRSvRecord(cnv, sampleId);
+                        CVRSvRecord record = svUtilities.makeCvrSvRecordFromGmlCnvIntragenicVariant(cnv, sampleId);
+                        try {
+                            svUtilities.populateEventInfoFromGmlCnvIntragenicVariant(record);
```
averyniceday, line 114: considering we are already making the CVRSvRecord from CnvIntrangenic, couldn't we call the populateeventinfo method from inside that function? seems like an extra step when it's all setting data members within the same class

cvr/src/test/java/org/cbioportal/cmo/pipelines/cvr/sv/CvrSvTestConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+
+package org.cbioportal.cmo.pipelines.cvr.sv;
+
+import org.springframework.context.annotation.*;
+
+@Configuration
+public class CvrSvTestConfiguration {
+
+    @Bean
+    public SvUtilities svUtilites() {
```
averyniceday, line 41: can we add this bean declaration in the CvrTestConfiguration and have it be the master file for initializing test beans?

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/sv/GMLSvDataReader.java
averyniceday, line 114: can we just call this from inside the `.makeCvrSvRecordFromGmlCnvIntragenicVariant` function? seems like this will always be coupled together?

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/sv/CVRSvDataReader.java
```diff
@@ -111,12 +112,17 @@ public void open(ExecutionContext ec) throws ItemStreamException {
             }
             reader.close();
         }
-        
         for (CVRMergedResult result : cvrData.getResults()) {
             String sampleId = result.getMetaData().getDmpSampleId();
             List<CVRSvVariant> variants = result.getSvVariants();
             for (CVRSvVariant variant : variants) {
-                CVRSvRecord record = new CVRSvRecord(variant, sampleId);
+                CVRSvRecord record = svUtilities.makeCvrSvRecordFromCvrSvVariant(variant, sampleId);
```
averyniceday, line 119: same comment as below about calling these populate `field` functions from inside the `.makeCvrSvRecordFromCvrSvVariant` since they will always be called together

averyniceday (changes requested):
Think we should call the populateEventInfo and related functions in the createSvRecord functions

averyniceday (approved):
lgtm

PR #967, opened 2022-08-18
Adjust oncokb-annotator script to process data_sv.txt

sheridancbio:


sheridancbio:
We got the go-ahead for using sv annotation from @zhx828 , so merging and deploying this change

PR #971, opened 2022-10-11
PDX Fixes

averyniceday:
- Assume validation passes
- Differentiate between cmo-arogs ci studies versus old bic studies
- Re-add metadata header indicator when first column is not used


sheridancbio:
This has been in production several months. This was to mitigate pdx merge issues.

PR #982, opened 2022-12-16
Az pipeline updates

callachennault:
Changes in `update-az-mskimpact.sh`:
- Standardize error reporting
- Filter files that are delivered
- Filter attributes in clinical files
- Add in clinical metadata headers
- Generate case lists


import-scripts/update-az-mskimpact.sh
```diff
@@ -46,7 +238,7 @@ fi
 # Create temporary directory to store data before subsetting
 if ! [ -d "$AZ_TMPDIR" ] ; then
     if ! mkdir -p "$AZ_TMPDIR" ; then
-        echo "Error : could not create tmp directory '$AZ_TMPDIR'" >&2
+        echo "ERROR: could not create tmp directory '$AZ_TMPDIR'" >&2
```
mandawilson, line 241: This should call report_error.

mandawilson (changes requested):

mandawilson (approved):
Excellent!

PR #986, opened 2022-12-30
Add age limit logic to AZ pipeline

sheridancbio:
This PR adds the `anonymize_age_at_seq_with_cap_py3.py` script. For the AZ MSKIMPACT pipeline, it is used as such: 
- Any `AGE_CURRENT` over 89 will be binned to >89.
- `AGE_AT_SEQ_REPORTED_YEARS` over 89 will be binned to >89.
- For any patient such that `sum(AGE_AT_SEQ_REPORTED_YEARS + OS_MONTHS) > 89`, `AGE_AT_SEQ_REPORTED_YEARS` will be capped to `>{max(0, 89 - OS_MONTHS)}`.
- `OS_MONTHS` will be altered to have a 2 digit precision limit.

Unit tests for this script are located at: `test-py3/test_anonymize_age_at_seq_with_cap.py`

The `anonymize_age_at_seq_with_cap_py3.py` script is now executed as part of the `update-az-mskimpact.sh` pipeline.


sheridancbio:
debugging and unit tests still needed, but overall logic is present

sheridancbio:
Some bug-fixing done. Script now runs. Output file has many blank values - investigate. Also a couple of TODO items left.

averyniceday (commented):
Just a general comment for future script development -- some of the methods in the LineProcessor class related to parsing files/headers are already available. I think it's okay for now since we want to push out AZ - but next time I think we should focus on using existing modules especially for shared functionality work

import-scripts/anonymize_age_at_seq_with_cap_py3.py
```diff
@@ -0,0 +1,421 @@
...
+        """
+        age_value = cols[age_col_index]
+        patient_id_value = cols[patient_id_col_index]
+
+        # Don't process blank age values
+        if not age_value:
+            return
+
+        # Handle age values that are already capped
+        if age_value[0] == '>':
```
sheridancbio, line 267: I think this also will behave appropriately below. So with 89 as an upper age limit, if we had ">80" as our current value and had os_years of 3 we would simply leave the value of ">80" and not change it. However, if os_years became 10 the sum will be 90 so we would be replacing the old ">80" value with (89 - 10) or ">79". When os_years became 20 the sum will be 100 so we would be replacing the old ">80" value with 89 - 20 or ">69"

import-scripts/anonymize_age_at_seq_with_cap_py3.py
```diff
@@ -0,0 +1,421 @@
...
+
+    # Transform the patient file
+    patient_file_reader = PatientFileWriter(
+        input_patient_file_path, output_patient_file_path, upper_age_limit, lower_age_limit, os_months_precision
+    )
+    patient_file_reader.write()
+
+    # Transform the clinical sample file
+    sample_file_writer = SampleFileWriter(
+        input_sample_file_path,
```
sheridancbio, line 415: The style of argument passing should match between here and the patient file reader above. If this is preferred then make the declaration of patient_file_reader match.

sheridancbio (commented):
Partial review (code only)

mandawilson (approved):
Great work!

import-scripts/anonymize_age_at_seq_with_cap_py3.py
```diff
@@ -0,0 +1,421 @@
...
+        patient_id_value = cols[patient_id_col_index]
+        age_current_value = cols[age_current_col_index]
+
+        # Don't process blank age values
+        if not age_current_value:
+            return
+
+        # Handle age values that are already capped
+        if age_current_value[0] == '>':
+            age_current_value = age_current_value[1:]
```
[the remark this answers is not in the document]
callachennault: @sheridancbio Yep - essentially if a value is already capped below the upper age limit (such as ">50" when the upper age limit is 89), there will be no change and will stay as ">50". I could modify the code to change it to >89 if that makes more logical sense.

import-scripts/anonymize_age_at_seq_with_cap_py3.py
callachennault: @sheridancbio Yes, this sounds correct to me! I have some unit tests that test these cases as well.

PR #998, opened 2023-01-21
az update repo/study name and do not filter unknown

sheridancbio:
- added "UNKNOWN" as a Mutation_Status to not be filtered
- adjusted references to repository base directory
- adjusted references to study directory

### DIFF
--- a/import-scripts/update-az-mskimpact.sh
+++ b/import-scripts/update-az-mskimpact.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 
-export AZ_DATA_HOME=$PORTAL_DATA_HOME/az-msk-impact
-export AZ_MSK_IMPACT_DATA_HOME=$AZ_DATA_HOME/mskimpact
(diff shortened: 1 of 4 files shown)


import-scripts/update-az-mskimpact.sh
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 
-export AZ_DATA_HOME=$PORTAL_DATA_HOME/az-msk-impact
-export AZ_MSK_IMPACT_DATA_HOME=$AZ_DATA_HOME/mskimpact
+export AZ_DATA_HOME="$PORTAL_DATA_HOME/az-data"
```
callachennault, line 3: Extremely small detail - but would we want to rename the repo to 'az_data' (with an underscore instead of a dash) to match the naming convention for the study?

callachennault (approved):

callachennault:
Thanks @sheridancbio, looks good! I left a small comment above

PR #999, opened 2023-01-22
determine the update-az-data execution by startday

sheridancbio:


callachennault (approved):
Thanks @sheridancbio!

PR #1001, opened 2023-01-25
add trigger tool logic for genie importer

sheridancbio:
- hostname now detected by import-tool (rather than being hardcoded)


callachennault (approved):
Looks good to me!

PR #1002, opened 2023-01-25
PDX Bug - Metadata Header Issue w/ Duplicating Columns

averyniceday:
Fix to clinicalfile utils - duplicate column function no longer assumes metadataheaders are present
Add a step to PDX pipeline to add metadata headers


sheridancbio (approved):
Looks good 👍

averyniceday:
For this use case I think it's fine - we're adding metadata headers after merging meaning we have no way to guaranteeing one set of values over another. In this case adding default values is acceptable (although a bit redundant).

PR #1004, opened 2023-01-31
Bugfix for cache clearing

averyniceday:
Script was not properly setting k8s config - conditional added to only point to public cluster for specific portal recaches

### DIFF
--- a/import-scripts/clear_cbioportal_persistence_cache.sh
+++ b/import-scripts/clear_cbioportal_persistence_cache.sh
@@ -75,12 +75,13 @@ if [ -z "$cluster_id" ] || [ -z "$deployment_id" ]; then
     exit 1
 fi
 
+unset KUBECONFIG_ARG
 if [ "$cluster_id" == "$CLUSTER_ID_DIGITS" ] ; then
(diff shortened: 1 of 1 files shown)


import-scripts/clear_cbioportal_persistence_cache.sh
```diff
@@ -75,12 +75,13 @@ if [ -z "$cluster_id" ] || [ -z "$deployment_id" ]; then
...
     /data/portal-cron/scripts/authenticate_service_account.sh
-fi
-unset KUBECONFIG_ARG
-if ! [ -z $PUBLIC_CLUSTER_KUBECONFIG ] ; then
-    KUBECONFIG_ARG="--kubeconfig $PUBLIC_CLUSTER_KUBECONFIG"
+else
+    if ! [ -z $PUBLIC_CLUSTER_KUBECONFIG ] ; then
+        KUBECONFIG_ARG="--kubeconfig $PUBLIC_CLUSTER_KUBECONFIG"
+    fi
 fi
```
inodb, line 85: `elif` more common in this case

inodb (approved):
Thanks!

PR #1008, opened 2023-02-02
Fix bug in changelog test that was producing different output from changelog generation script

callachennault:
The unit test code for AZ changelog generation script was parsing the git diff output differently than the changelog script itself, causing unit test output to differ from actual script output. The unit tests now call the same git diff parsing functions and produce consistent output.


averyniceday (approved):
lgtm, but maybe we should rename variables for readability in the future

PR #1011, opened 2023-02-10
Update number of arguments passed into getAnnotatedRecordsUsingPOST call

callachennault:
The number of arguments accepted by the `getAnnotatedRecordsUsingPOST` function was updated in this PR: https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/233. All calls to this function have been updated to be consistent with this change.


sheridancbio:
Essentially the same changes were made as part of #1061 which has been merged to master.

PR #1012, opened 2023-02-13
Add Gene Panel Matrix File Standardization

averyniceday:
Bug becomes apparent for studies which merge unlinked Archer samples (e.g. Lehigh)

- update validation util standardize gene panel matrix function
-  wrapper script for calling above function
- add calls inside dmp pipeline to standardize after subsetting

### DIFF
--- a/import-scripts/standardize_gene_matrix_file.py
+++ b/import-scripts/standardize_gene_matrix_file.py
@@ -1,0 +1,33 @@
+import sys
+import argparse
+import os
+
+import validation_utils
+
+ERROR_FILE = sys.stderr
+OUTPUT_FILE = sys.stdout
+
+def main():
+    # get command line stuff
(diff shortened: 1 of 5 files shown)


import-scripts/standardize_gene_matrix_file.py
```diff
@@ -0,0 +1,33 @@
+import sys
```
callachennault, line 1: Could be useful to add a small comment to the beginning of the script explaining what the script does. I added this to the standardize_clinical_data script from a few weeks ago: https://github.com/knowledgesystems/cmo-pipelines/blob/dc9547064a24d06a0cafb5cf3fe838269f2fd860/import-scripts/standardize_clinical_data.py#L1-L12

callachennault (approved):
Looks good to me @averyniceday!

PR #1014, opened 2023-02-16
Standardize blank values in CNA file to NA

callachennault:
Import of az_mskimpact is failing with public import tools in part due to blank values in the `data_CNA.txt` file. This PR introduces the `standardize_cna_data.py` script and calls it on each update of the az_mskimpact dataset.

### DIFF
--- a/import-scripts/standardize_cna_data.py
+++ b/import-scripts/standardize_cna_data.py
@@ -17,13 +17,19 @@ import os
 import validation_utils
 
 ERROR_FILE = sys.stderr
-OUTPUT_FILE = sys.stdout
 
 
 def main():
     # Receive path to clinical file as an argument
     parser = argparse.ArgumentParser(prog='standardize_cna_data.py')
-    parser.add_argument('-f', '--filename', dest='filename', action='store', required=True, help='path to data_CNA.txt file')
+    parser.add_argument(
+        '-f',
+        '--filename',
+        dest='filename',
+        action='store',
+        required=True,
+        help='path to data_cna.txt file',
+    )
 
     args = parser.parse_args()
     filename = args.filename
--- a/import-scripts/standardize_cna_data.py
+++ b/import-scripts/standardize_cna_data.py
@@ -1,0 +1,45 @@
+#! /usr/bin/env python
+
+""" standardize_cna_data.py
+This script rewrites a CNA file to replace all NA placeholders/blank values with 'NA'.
+
+Usage:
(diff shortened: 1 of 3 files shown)


import-scripts/standardize_cna_data.py
```diff
@@ -0,0 +1,45 @@
+#! /usr/bin/env python
```
sheridancbio, line 1: I think it is standard to have no space after the exclamation point

import-scripts/standardize_cna_data.py
```diff
@@ -0,0 +1,45 @@
...
+
+""" standardize_cna_data.py
+This script rewrites a CNA file to replace all blank values with 'NA'.
+
+Usage:
+    python standardize_cna_data.py --filename $INPUT_CNA_FILE
+Example:
+    python standardize_cna_data.py --filename /path/to/data_CNA.txt
```
sheridancbio, line 9: our standard filename pattern is now 'data_cna.txt' for the cna file. I know that we use data_CNA.txt for the (legacy) dmp studies, but perhaps we will move those to the standard at some point. So I would recommend we use the standard filename 'data_cna.txt' in this file. It also makes me wonder if we should standardize our filenames in the delivered study but that's out of scope here.

import-scripts/update-az-mskimpact.sh
```diff
@@ -246,6 +246,17 @@ function standardize_clinical_data() {
     mv "$SAMPLE_OUTPUT_FILEPATH" "$SAMPLE_INPUT_FILEPATH"
 }
 
+function standardize_cna_data() {
+    DATA_CNA_INPUT_FILEPATH="$AZ_MSK_IMPACT_DATA_HOME/data_CNA.txt"
+    DATA_CNA_OUTPUT_FILEPATH="$AZ_MSK_IMPACT_DATA_HOME/data_CNA.txt.standardized"
+
+    # Standardize the CNA file to use NA for blank values
+    $PYTHON_BINARY $PORTAL_HOME/scripts/standardize_cna_data.py -f "$DATA_CNA_INPUT_FILEPATH" > "$DATA_CNA_OUTPUT_FILEPATH" &&
```
sheridancbio, line 254: I'm not certain what the trailing '&&" does ... if this is to connect with the lower "mv" command then I would not have the blank lines between and would double indent the lower line

sheridancbio (commented):
looks good. I made a few suggestions. The trailing "&&" might be an error though .. I am not sure

PR #1017, opened 2023-02-20
Add another sample id header (SV post processed for AZ)

averyniceday:
Without this check, assumes first column is not the header and adds it to case list


callachennault (approved):
Looks good to me!

PR #1019, opened 2023-02-20
Correctly handle no change to clinical files in changelog generation script

callachennault:
Changelog generation script fails when there is no git diff to parse. This PR fixes that bug and now runs successfully even if the clinical files haven't changed.


mandawilson (approved):
Great!

PR #1020, opened 2023-03-03
Move age at sequencing calculation from CVR to DDP

callachennault:


ddp/ddp_pipeline/src/main/java/org/mskcc/cmo/ks/ddp/pipeline/DDPSeqDateTasklet.java
```diff
@@ -123,6 +130,7 @@ public Map<String, Date> getFirstSeqDatePerPatientFromFile(String filename, Buff
                 LOG.warn(warnings.get(w)); 
             }
         }
-        return patientFirstSeqDateMap;
+        DDPUtils.setSampleSeqDateMap(sampleSeqDateMap);
+        DDPUtils.setPatientFirstSeqDateMap(patientFirstSeqDateMap);
```
sheridancbio, line 134: we are setting a class data members in a remote class in these two function calls. This is equivalent to using a global variable to pass data around in a program - somewhat error prone but not invalid. Also, we use DDPUtils to hold several other of these global mappings ... so that they can later be used for formatting records I believe, so this is in the current pattern of usage. So ok.

ddp/ddp_pipeline/src/main/java/org/mskcc/cmo/ks/ddp/pipeline/util/DDPUtils.java
```diff
@@ -166,6 +174,43 @@ public static String resolvePatientCurrentAge(DDPCompositeRecord compositeRecord
...
+    public static String resolveAgeAtSeqDate(String sampleId, String patientBirthDate) throws ParseException {
+        Long birthDateInDays = getDateInDays(patientBirthDate);
+        Long referenceDateInDays = getDateInDays(sampleSeqDateMap.get(sampleId));
+        // if either date is null do not calculate age at sequencing date
+        Double age = (referenceDateInDays == null || birthDateInDays == null) ? null : (referenceDateInDays - birthDateInDays) / (DAYS_TO_YEARS_CONVERSION);
+        if (age == null) {
+            // log cases where AGE_AT_SEQ_REPORTED_YEARS is null because required values are null
+            String osMonthsLogMessage = getAgeAtSeqDateLogging(sampleId, String.valueOf(age), referenceDateInDays);
+            LOG.debug(osMonthsLogMessage);
+        }
```
sheridancbio, line 204: If a
`return null`
is added in the if body then we could change "else if" into plain "if" below and also drop the ternary operator at the end of this function (if we also put a `return null` at then end of the (age < 0) conditional block)

ddp/ddp_pipeline/src/main/java/org/mskcc/cmo/ks/ddp/pipeline/util/DDPPatientListUtil.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Memorial Sloan-Kettering Cancer Center.
+ * Copyright (c) 2018-2023 Memorial Sloan-Kettering Cancer Center.
```
sheridancbio, line 2: It's not clear whether we have standardized on a having spaces around the hyphen between the two years at the top of every file ... but I think that would be a good standard. (there is another no space instance in a later file)

ddp/ddp_pipeline/src/main/java/org/mskcc/cmo/ks/ddp/pipeline/util/DDPUtils.java
```diff
@@ -166,6 +174,43 @@ public static String resolvePatientCurrentAge(DDPCompositeRecord compositeRecord
         return anonymizePatientAge(age.intValue());
     }
 
+    public static String getAgeAtSeqDateLogging(String sampleId, String ageAtSeqDate, Long referenceInDays) {
```
sheridancbio, line 177: Because this is only called when AgeAtSeqDate is null, perhaps the function name should be getAgeAtSeqDateNullLogging() ?
Otherwise it sounds like a general purpose log string generator for any sample

ddp/ddp_pipeline/src/main/java/org/mskcc/cmo/ks/ddp/pipeline/util/DDPUtils.java
```diff
@@ -166,6 +174,43 @@ public static String resolvePatientCurrentAge(DDPCompositeRecord compositeRecord
...
+     * the patient date of birth or the date of sequencing is null this returns null.
+     *
+     * @param sampleId
+     * @param patientBirthDate String date of birth
+     * @return age in years at sequencing, null if we don't know patient date of birth or sequencing date
+     * @throws ParseException
+     */
+    public static String resolveAgeAtSeqDate(String sampleId, String patientBirthDate) throws ParseException {
+        Long birthDateInDays = getDateInDays(patientBirthDate);
+        Long referenceDateInDays = getDateInDays(sampleSeqDateMap.get(sampleId));
```
sheridancbio, line 197: Is this always the sample sequencing date? If so, I suggest an equally concrete name for this variable as "birthDateInDays" .. such as sampleSeqDateInDays

ddp/ddp_pipeline/src/test/java/org/mskcc/cmo/ks/ddp/pipeline/DDPSeqDateTaskletTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 Memorial Sloan-Kettering Cancer Center.
+ * Copyright (c) 2019-2023 Memorial Sloan-Kettering Cancer Center.
```
sheridancbio, line 2: spaces around hyphen between years

ddp/ddp_pipeline/src/test/java/org/mskcc/cmo/ks/ddp/pipeline/util/DDPUtilsTest.java
```diff
@@ -268,68 +297,6 @@ public void resolveOsMonthsTest() throws ParseException {
...
-     *      by the DDP pipeline yet. At this time there is no need to implement the
-     *      DDP equivalent of the darwin/darwin_age.txt file as it is still being
-     *      generated by the Darwin pipeline.
-     *
-     *      The Darwin pipeline calculates "AGE" as the years since birth:
-     *              (currentYear - patientBirthYear)
-     *      This behavior is mirrored in DDPUtils.resolvePatientAgeAtDiagnosis() and
-     *      will not be modified until further notice.
-     *
-     *      Eventually the age of diagnosis will be calculated on a per-sample basis, using
```
sheridancbio, line 284: I remember this discussion. There were subtleties about defining when a patient was first diagnosed because a patient may have several different/independent cancer diagnosis. We had talked about grouping samples based on the type of cancer and possibly having multiple first diagnosis dates. Maybe we are no longer going to try to do this in the future - and to only report the very first day a patient had *any* sample sequenced. But I think maybe we should ask whether a note about this intention should be made anywhere (perhaps on a card or a roadmap planning document) Maybe @n1zea144 can comment here.

ddp/ddp_source/src/main/java/org/mskcc/cmo/ks/ddp/source/composite/DDPCompositeRecord.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2019 Memorial Sloan-Kettering Cancer Center.
+ * Copyright (c) 2018-2023 Memorial Sloan-Kettering Cancer Center.
```
sheridancbio, line 2: spaces around hyphen

import-scripts/dmp-import-vars-functions.sh
```diff
@@ -195,13 +195,16 @@ function import_mskimpact_supp_date_to_redcap {
...
+    if ! import_project_to_redcap $MSK_IMPACT_DATA_HOME/data_clinical_ddp_age_at_seq.txt mskimpact_clinical_ddp_age_at_seq ; then return_value=1 ; fi
     if ! import_project_to_redcap $MSK_IMPACT_DATA_HOME/data_clinical_ddp_pediatrics.txt mskimpact_data_clinical_ddp_demographics_pediatrics ; then return_value=1 ; fi
     if ! import_project_to_redcap $MSK_IMPACT_DATA_HOME/data_timeline_ddp_chemotherapy.txt mskimpact_timeline_chemotherapy_ddp; then return_value=1 ; fi
     if ! import_project_to_redcap $MSK_IMPACT_DATA_HOME/data_timeline_ddp_radiation.txt mskimpact_timeline_radiation_ddp ; then return_value=1 ; fi
     if ! import_project_to_redcap $MSK_IMPACT_DATA_HOME/data_timeline_ddp_surgery.txt mskimpact_timeline_surgery_ddp ; then return_value=1 ; fi
     return $return_value
 }
 
+
+
```
sheridancbio: Only 1 blank line?

sheridancbio (approved):
Looks good. Just suggestions/comments and whitespace adjustment requests.

ddp/ddp_pipeline/src/main/java/org/mskcc/cmo/ks/ddp/pipeline/AgeAtSeqDateWriter.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public void close() throws ItemStreamException {
+        flatFileItemWriter.close();
+    }
+
+    @Override
+    public void write(List<? extends CompositeResult> compositeResults) throws Exception {
+        List<String> records = new ArrayList<>();
+        for (CompositeResult result : compositeResults) {
+            if (result.getAgeAtSeqDateResults() == null || result.getAgeAtSeqDateResults().isEmpty()) {
```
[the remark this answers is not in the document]
callachennault: The fetch of the clinical data in ddp is re-attempted for all patients every day - marking as resolved

ddp/ddp_pipeline/src/test/java/org/mskcc/cmo/ks/ddp/pipeline/util/DDPUtilsTest.java
callachennault: `resolvePatientAgeAtDiagnosis()` function was removed in this PR because it's not called anywhere (only in this test file). Marking resolved

PR #1027, opened 2023-03-27
Add path_score column to mutation file and update germline filtering code

callachennault:
This PR:

- Reverts the code change from this PR ([https://github.com/knowledgesystems/cmo-pipelines/pull/1013](https://github.com/knowledgesystems/cmo-pipelines/pull/1013)) which appended `PATH_SCORE` to the `MUTATION_STATUS` column in the `data_mutations_extended.txt` file
- Adds `PATH_SCORE` as its own column in the `data_mutations_extended.txt` file
- Updates how the `filter_non_somatic_events_py3.py` script filters non-somatic events by checking the `PATH_SCORE` column for each germline event and removing any that do not have a path score of 'Pathogenic' or 'Likely Pathogenic'
- Adds unit tests for the `filter_non_somatic_events_py3.py` script


averyniceday (approved):
lgtm

PR #1028, opened 2023-03-27
fix genie import script

sheridancbio:
- manage trigger files properly
- add an inhibit period after emailing failure notices for db mismatch

### DIFF
--- a/import-scripts/import-genie-data.sh
+++ b/import-scripts/import-genie-data.sh
@@ -19,6 +19,28 @@ FLOCK_FILEPATH="/data/portal-cron/cron-lock/import-genie-data.lock"
         exit 1
     fi
 
+    # inhibit import attempt if db mismatch occurred recently
+    inhibit_counter=0
+    if [ -f "$GENIE_IMPORT_VERSION_EMAIL_INHIBIT_FILENAME" ] ; then
+        unset inhibit_counter_line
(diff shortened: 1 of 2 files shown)


import-scripts/import-genie-data.sh
```diff
@@ -131,13 +158,19 @@ FLOCK_FILEPATH="/data/portal-cron/cron-lock/import-genie-data.lock"
 
     # send email if db version isn't compatible
     if [ $DB_VERSION_FAIL -gt 0 ]; then
-        EMAIL_BODY="The genie database version is incompatible. Imports will be skipped until database is updated."
+        EMAIL_BODY="The genie database version is incompatible. Imports will be skipped until database is updated, and file $GENIE_IMPORT_VERSION_EMAIL_INHIBIT_FILENAME has been created to inhibit imports for the next "$VERSION_MISMATCH_EMAIL_INHIBIT_COUNTER_START" minutes."
```
mandawilson, line 161: Please remove the inside quotes.

mandawilson (approved):

import-scripts/import-genie-data.sh
sheridancbio: fixed

sheridancbio:
needed to move inhibit counter code block higher in the script so that it is decremented even if there is no import trigger file present.

PR #1031, opened 2023-04-03
Restore add-age-at-seq-report for genie and update calculation logic

callachennault:
This PR:

- Reintroduces the `add-age-at-seq-report.py` script and its corresponding call in `subset-impact-data.sh` for genie. This code had previously been removed in this PR ([https://github.com/knowledgesystems/cmo-pipelines/pull/1020](https://github.com/knowledgesystems/cmo-pipelines/pull/1020)) but we still need it in order to add the `AGE_AT_SEQ_REPORT` column to genie data.
- Modifies the `add-age-at-seq-report.py` script to calculate `AGE_AT_SEQ_REPORT` by reading the `AGE_AT_SEQ_REPORTED_YEARS` attribute from the clinical sample file and converting it to days. This fixes the bug that was occurring in the prior method of calculation (described in more detail in the Age at Sequencing ADR).

### DIFF
--- a/import-scripts/add-age-at-seq-report.py
+++ b/import-scripts/add-age-at-seq-report.py
@@ -1,0 +1,150 @@
+import sys
+import optparse
+import csv
+import os
+import math
+from datetime import datetime
+
+ERROR_FILE = sys.stderr
+OUTPUT_FILE = sys.stdout
+
+# globals
+AGE_AT_SEQ_REPORT_FIELD = 'AGE_AT_SEQ_REPORT'
+AGE_FIELD = 'AGE'
+SEQ_DATE_FIELD = 'SEQ_DATE'
+AVERAGE_DAYS_PER_YEAR = 365.2422
+
+PATIENT_SAMPLE_MAP = {}
+SAMPLE_SEQ_DATE_MAP = {}
+PATIENT_AGE = {}
+
+
+def load_sample_seq_date(seq_date_file, convert_to_days):
+	""" Loads SEQ_DATE from seq date file provided. """
+	data_file = open(seq_date_file, 'rU')
+	data_reader = csv.DictReader(data_file, dialect = 'excel-tab')
+
+	for line in data_reader:
+		# skip samples not found in the clinical file
+		if not line['SAMPLE_ID'] in PATIENT_SAMPLE_MAP.keys():
+			continue
+		seq_date = datetime.strptime(line['SEQ_DATE'], '%a, %d %b %Y %H:%M:%S %Z')
+		difference_in_years = int(math.ceil(((datetime.now() - seq_date).days)/AVERAGE_DAYS_PER_YEAR))
+		try:
+			age_at_seq_report_value = (PATIENT_AGE[line['PATIENT_ID']] - difference_in_years)
+			if convert_to_days:
+				age_at_seq_report_value = age_at_seq_report_value*AVERAGE_DAYS_PER_YEAR
+			SAMPLE_SEQ_DATE_MAP[line['SAMPLE_ID']] = str(int(math.ceil(age_at_seq_report_value)))
+		except KeyError:
+			print "Patient age not found for '" + line['PATIENT_ID'] + "'"
+			SAMPLE_SEQ_DATE_MAP[line['SAMPLE_ID']] = 'NA'
+
+
+def load_patient_age(age_file):
+	""" Loads AGE from age file. """
+	data_file = open(age_file, 'rU')
+	data_reader = csv.DictReader(data_file, dialect = 'excel-tab')
+
+	for line in data_reader:
+		# skip records not found in clinical file 
+		if not line['PATIENT_ID'] in PATIENT_SAMPLE_MAP.values():
+			continue
+		PATIENT_AGE[line['PATIENT_ID']] = int(line['AGE'])
+
+
+def load_patient_sample_mapping(clinical_file):
+	""" Loads patient-sample mapping. """
+	data_file = open(clinical_file, 'rU')
+	data_reader = csv.DictReader(data_file, dialect = 'excel-tab')
+	for line in data_reader:
+		PATIENT_SAMPLE_MAP[line['SAMPLE_ID']] = line['PATIENT_ID']
+	data_file.close()
+
+
+def add_age_at_seq_report(clinical_file):
+	header = get_file_header(clinical_file)
+
+	# add age at seq report column in not already present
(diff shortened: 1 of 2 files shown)


averyniceday (approved):
Changes look good to me - couple suggestions
1) Rename the script argument/flags - clinical sample file/clinical output file or something of the sort
2) Check if the required headers are in the file to avoid NPE

import-scripts/add-age-at-seq-report.py
```diff
@@ -1,55 +1,49 @@
...
-	data_file = open(age_file, 'rU')
-	data_reader = csv.DictReader(data_file, dialect = 'excel-tab')
-
-	for line in data_reader:
-		# skip records not found in clinical file 
-		if not line['PATIENT_ID'] in PATIENT_SAMPLE_MAP.values():
-			continue
-		PATIENT_AGE[line['PATIENT_ID']] = int(line['AGE'])
+				age_at_seq_report_value = int(age_at_seq_reported_years) * AVERAGE_DAYS_PER_YEAR
+			SAMPLE_AGE_AT_SEQ_REPORT_MAP[sample_id] = str(int(math.floor(age_at_seq_report_value)))
```
sheridancbio, line 86: I notice that ceil() was changed to floor() here. That may be fine ... so long as the same calculation is being performed consistently (for age at sequencing and age at death)

import-scripts/add-age-at-seq-report.py
```diff
@@ -1,55 +1,49 @@
...
+		except ValueError:
+			print AGE_AT_SEQ_REPORTED_YEARS_FIELD + " not found for '" + sample_id + "'"
```
sheridancbio, line 88: 👍

import-scripts/subset-impact-data.sh
```diff
@@ -103,9 +103,9 @@ if [ $STUDY_ID == "genie" ]; then
 
         # add age at seq report using cvr/seq_date.txt
         echo "Adding AGE_AT_SEQ_REPORT to $OUTPUT_DIRECTORY/data_clinical_supp_sample.txt"
-        $PYTHON_BINARY $PORTAL_SCRIPTS_DIRECTORY/add-age-at-seq-report.py --clinical-file="$OUTPUT_DIRECTORY/data_clinical_supp_sample.txt" --seq-date-file="$INPUT_DIRECTORY/cvr/seq_date.txt" --age-file="$INPUT_DIRECTORY/ddp/ddp_age.txt" --convert-to-days="true"
+        $PYTHON_BINARY $PORTAL_SCRIPTS_DIRECTORY/add-age-at-seq-report.py --clinical-file="$OUTPUT_DIRECTORY/data_clinical_supp_sample.txt" --sample-file="$INPUT_DIRECTORY/data_clinical_sample.txt" --convert-to-days="true"
```
sheridancbio: if you define a symbolic variable for the sample file path, you could use it here and also on line 108. That would help avoid a typo difference between the two lines or a maintenance mistake later (changing this line and forgetting to propagate it to 108). It also looks like such a variable could be used in other places in this script.

sheridancbio (approved):
👍

import-scripts/add-age-at-seq-report.py
callachennault: Thanks for noticing this @sheridancbio! I changed it from `ceil()` to `floor()` because it seemed to me that rounding up this value would introduce the same kind of errors that we were seeing before. When I run this same code but with the `math.ceil()` function, I see 444 records where the patient's `age_at_sequencing_reported` is greater than `int_dod`, and in a lot of these cases `AGE_AT_SEQUENCING_REPORTED` is only greater by one day. When I run it with `math.floor()`, I only see 25 records where `AGE_AT_SEQUENCING_REPORTED` is greater than `INT_DOD`. I guess it depends what we want. I'm not sure how this value is being converted to years in the genie portal - ie if they're dividing by 365 or 365.2422 (the multiplier we're using). It could yield different results based on whether we're using `floor()` or `ceil()`.

To your second point - that we need the calculations to be performed consistently - I don't think this would be possible without a larger code change. Age at death (INT_DOD) is calculated [here](https://github.com/knowledgesystems/cmo-pipelines/blob/72c387529b737952fba0d551a5ec0bb9f7d58149/ddp/ddp_pipeline/src/main/java/org/mskcc/cmo/ks/ddp/pipeline/model/SuppVitalStatusRecord.java#L63), where it calls [this function](https://github.com/knowledgesystems/cmo-pipelines/blob/4ea50c11a13b18e5601ec53877a9c634715fe31c/ddp/ddp_pipeline/src/main/java/org/mskcc/cmo/ks/ddp/pipeline/util/DDPUtils.java#L208-L216). From reading the code it seems that `INT_DOD` is precise to the day (and might be revealing the patient's birthdate?) Maybe we could discuss this a bit?

sheridancbio: I guess it is a little confusing and philosophical to decide exactly when a person changes from being X years old to being X + 1 years old. Do you become a year older at midnight of the day of your birth, at midnight of the day after your birth, or at your time of birth on the day of your birth? Even the last option is debatable because a year of time (astronomically) includes a fraction of a day, and that fraction is what we are either rounding up or down here in an aggregate over a person's lifetime. I think your comment about knowing the way that genie reverses this conversion is on point - it does seem to me that it would be improper for us to send to genie a number of days for a patient whom we consider to be 30 ... and have genie integrate that patient into their dataset with an age of 31. It seems to me that whatever (year) age we consider the patient to be we should convert that to days in a way where the reverse conversion by genie will always result in the original age we had.

If our age at death calculation, I see the same conversion factor here also present in resolvePatientCurrentAge(), but the reference you give points to resolveIntervalInDays() and that only does anonymization of age based on upper/lower limit rather than on reducing the granularity to years. So I think something is missing here ... the granularity of the code you pointed me to is days (I think) .. but isn't age at death recorded in years in our clinical file? There must be another step where we convert the days into years, and that would be the logic which should ideally be consistent I guess.

callachennault: @sheridancbio Here's the code where genie converts the age at seq in days to years: https://github.com/Sage-Bionetworks/Genie/blob/53db39ccb6e423743df49c1f0e115c9a7576af69/genie/database_to_staging.py#L1132-L1137. I've modified our code to do the reverse calculation. I'll note that this increases the number of cases where age of sequencing is greater than INT_DOD to 636 records but again it seems that the majority of these cases are off by 1 day, and now that we're doing the exact reverse calculation they should end up with the correct year value.

PR #1038, opened 2023-05-23
Modify merge script to ignore blank lines in subset and excluded sample list files

callachennault:
We ran into a bug with the merge script when a blank line in the subset list file was stored in the reference set as an empty string. The comparison looked for matches against any of the data values in the row. Any empty string in the data row matched the reference set and extraneous rows were kept. We fixed this issue by ignoring blank lines in the subset and excluded list files, but the comparison can still be risky as it can match against any value in the row. We added a note to the script explaining possible issues this can cause since we're not using a specific key column to compare to the reference set.


sheridancbio (approved):
Good find and good fix. The comment is on-target - we should as quickly as possible correct this so that the identifiers which are references for inclusion/exclusion should only be compared against the columns in which the values are expected to be present.

PR #1039, opened 2023-05-25
SMILE Connection Error Handling

averyniceday:
Original Bug
---------
In the event a NATS server connection could not be established, an exception would be thrown due to an inability to create the bean and entire pipeline would fail

Fix
---------
Catch the exception and continue
Update the SmileTasklet to skip publishing if connection is not established 

TODO/???
----------
Bean does not shutdown gracefully due to how the `shutdown()` method is implemented [here](https://github.com/mskcc/smile-messaging-java/blob/master/src/main/java/org/mskcc/cmo/messaging/impl/JSGatewayImpl.java#L204) 

Spring automatically calls the `shutdown()` method for beans and an exception is being thrown. However, this happens after the job completes on application shutdown. Might be okay to merge and leave this for a future fix?


callachennault (approved):
@averyniceday Looks good, small typo fix needed

PR #1044, opened 2023-06-20
Update scripts to import into cgds_hg1938 database

callachennault:
Add functionality to allow importing into new hg19/hg38 DB and portal


sheridancbio (approved):
👍

PR #1050, opened 2023-06-28
updates to persistence cache and kubeconfig for new EKS cluster

averyniceday:
- persistence cache, new version of redis in EKS cluster uses different syntax
- kubeconfig name - for interacting with new cluster when clearing cache


callachennault (approved):
looks good @averyniceday

PR #1052, opened 2023-07-11
move deployed annotationPipeline jar to lib directory

sheridancbio:
this decouples the development area in the cmo-pipelines git repository clone from the production area where in-production compiled jar files are (/data/portal-cron/lib).


averyniceday (commented):
does the build script auto-move the annotationPipeline jar to the lib directory?

averyniceday (approved):

PR #1055, opened 2023-07-18
Remove non signed out filtering from az script

callachennault:
PR Closed on 9/25 because germline filtering currently works as intended


sheridancbio:
These changes are not in production. I'm trying to close out recent cmo-pipelines PRs -- are these changes still under consideration, or is this no longer being pursued?

PR #1056, opened 2023-07-27
add nightly automated import of cmo-access

sheridancbio:
New import script which fetches cmo-access datasource, extracts patient list, fetches ddp timeline data, and imports study into msk-portal

### DIFF
--- a/import-scripts/fetch-ddp-and-import-cmo-access-data.sh
+++ b/import-scripts/fetch-ddp-and-import-cmo-access-data.sh
@@ -1,0 +1,262 @@
+#!/bin/bash
+
+# 0. acquire the exclusive file lock
+# 1. reset cmo-access git repository to match the upstream main branch (using similar lagic to
+#    the pipelines importer --fetch-data, but in script because of the different branch names.
+# 2. extract a list of dmp patients in the study (only "P-" prefixed)
+# 3. query the ddp server for associated timeline/clinical information on these patients
+# 4. create meta files for the 3 timeline data files and copy timeine info into the data repository
+# 5. fix case list files by overwriting stable_id and cancer_study_identifier
+# 6. check importer compatibility with the database schema version
+# 7. import the study to the msk-portal
+# 8. clear the persistence cache for the msk-portal
+# 9. send email notification of study import
+# 10. commit changes/updates into the data repository clone
+# 11. push changeset to the data repository
+
+MY_FLOCK_FILEPATH="/data/portal-cron/cron-lock/fetch-ddp-and-import-cmo-access-data.lock"
+(
+    echo $(date)
+
+    # check lock so that executions of this script not overlap
+    if ! flock --nonblock --exclusive $my_flock_fd ; then
+        echo "Failure : could not acquire lock for $MY_FLOCK_FILEPATH -- the fetch from yesterday is still running."
+        exit 1
+    fi
+
+    source $PORTAL_HOME/scripts/dmp-import-vars-functions.sh
+    source $PORTAL_HOME/scripts/clear-persistence-cache-shell-functions.sh
+
+    PIPELINES_EMAIL_LIST="cbioportal-pipelines@cbioportal.org"
+    CANCER_STUDY_IDENTIFIER="mixed_msk_cfdna_research_access"
+    IMPORT_SYMLINK_FILEPATH="/data/portal-cron/cbio-portal-data/cmo-access/mixed_msk_cfdna_research_access"
+    now=$(date "+%Y-%m-%d-%H-%M-%S")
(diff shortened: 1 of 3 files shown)


import-scripts/fetch-ddp-and-import-cmo-access-data.sh
```diff
@@ -0,0 +1,262 @@
...
+        exit 1
+    fi
+
+    # create tmpdir if it does not exist, and wipe contents
+    if [[ -d "$CMO_ACCESS_TMPDIR" ]] ; then
+        if [[ "$CMO_ACCESS_TMPDIR" != "/" ]] ; then
+            rm -rf "$CMO_ACCESS_TMPDIR"/*
+        fi
+    else
+        mkdir "$CMO_ACCESS_TMPDIR"
```
averyniceday, line 61: where's CMO_ACCESS_TMPDIR defined?

import-scripts/fetch-ddp-and-import-cmo-access-data.sh
```diff
@@ -0,0 +1,262 @@
...
+    # create tmpdir if it does not exist, and wipe contents
+    if [[ -d "$CMO_ACCESS_TMPDIR" ]] ; then
+        if [[ "$CMO_ACCESS_TMPDIR" != "/" ]] ; then
+            rm -rf "$CMO_ACCESS_TMPDIR"/*
+        fi
+    else
+        mkdir "$CMO_ACCESS_TMPDIR" 
+    fi
+
+    function fetch_git_repo_data_updates() {
```
averyniceday, line 64: can we instead update the pipelines code to recognized access as a fetch - see here: 

https://github.com/knowledgesystems/cmo-pipelines/blob/master/import-scripts/fetch-dmp-data-for-import.sh#L115

import-scripts/fetch-ddp-and-import-cmo-access-data.sh
```diff
@@ -0,0 +1,262 @@
...
+
+    function fetch_ddp_demographics_and_timeline() {
+        rm -f "$cmo_access_dmp_pids_filepath"
+        rm -f "$cmo_access_seq_date_filepath"
+        rm -rf "$cmo_access_ddp_output_dirpath"
+        mkdir "$cmo_access_ddp_output_dirpath"
+        echo "SAMPLE_ID	PATIENT_ID	SEQ_DATE" > "$cmo_access_seq_date_filepath" # create placeholder seq_date file
+        egrep -v "^#" $CMO_ACCESS_DATA_HOME/data_clinical_patient.txt | awk -F'\t' 'NR==1 { for (i=1; i<=NF; i++) { f[$i] = i } }{ if ($f["PATIENT_ID"] != "PATIENT_ID") { print $(f["PATIENT_ID"]) } }' | egrep "^P-" | sort | uniq > $cmo_access_dmp_pids_filepath
+        CMO_ACCESS_DDP_DEMOGRAPHICS_RECORD_COUNT=$(wc -l < $cmo_access_dmp_pids_filepath)
+        if [ $CMO_ACCESS_DDP_DEMOGRAPHICS_RECORD_COUNT -le $DEFAULT_DDP_DEMOGRAPHICS_ROW_COUNT ] ; then
```
averyniceday, line 118: where is DEFAULT_DDP_DEMOGRAPHICS_ROW_COUNT defined? is this a check to make sure we don't have no records?

import-scripts/fetch-ddp-and-import-cmo-access-data.sh
sheridancbio: good catch. I did not check in my changes to automation_environment.sh ... this is defined there. I will add this to the branch.

import-scripts/fetch-ddp-and-import-cmo-access-data.sh
sheridancbio: The difficulty here is that the branch name for the working directory in cmo-access is "main", but our pipelines fetch routine hardcodes the "master" branch as the one to be fetched. I agree that we should update pipelines to work for this repository, so some command line argument should be allowed for the working branch to be specified. However, currently we cannot build the pipelines codebase because of the java11 issue. As soon as we can build pipelines I will make a PR to add the working branch command line argument (or add a column in the data-sources sheet on portal-importer-configuration sheet if that is easier).

import-scripts/fetch-ddp-and-import-cmo-access-data.sh
sheridancbio: This environment variable is defined in dmp-import-vars-functions.sh as "2". I think there is some error which occurs if you query for just a single ddp fetch, so we always want a minimum of 2 patients when doing a ddp fetch. I don't think I have made any modifications to dmp-import-vars-functions.sh.

averyniceday (approved):

PR #1060, opened 2023-08-08
Add cvr rest api timeout retry

sheridancbio:
This PR contains two stages:
- stage 1 : updates to the codebase to make it compatible with java 11 compilation (to use the latest cBioPortal codebase)
- stage 2 : updates to employ a standard timeout/retry component in 7 classes which use RestTemplate API calls


common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpClientWithTimeoutAndRetry.java
```diff
@@ -0,0 +1,173 @@
...
+                lastResponseBodyStringAfterException = null; // reset before each request
+                numberOfRequestsAttempted = numberOfRequestsAttempted + 1;
+                lastTimeoutUsed = timeoutProgression.getNextTimeoutForRequest();
+		        requestFactory.setConnectTimeout(lastTimeoutUsed);
+		        requestFactory.setReadTimeout(lastTimeoutUsed);
+                ResponseEntity<T> response = restTemplate.exchange(url, method, requestEntity, returnType, uriVariables);
+                if (response.getStatusCode().is5xxServerError()) {
+                    pauseForMilliseconds(lastTimeoutUsed);
+System.out.println("got error: " + response.getStatusCode().toString());
+                    pauseForMilliseconds(lastTimeoutUsed);
```
n1zea144, line 139: Did you intend for this additional pause?

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpClientWithTimeoutAndRetry.java
```diff
@@ -0,0 +1,173 @@
...
+    public boolean exceptionCausedByTimeout(RestClientException e) {
+        return e.getRootCause() instanceof SocketTimeoutException && e.getRootCause().getMessage().toLowerCase().contains("timed out");
+    }
+    /* the exchange method takes similar arguments to HttpTemplate.exchange and it sets the intial timeout from data member initialTimeout.
+     * if the timeout is reached (either for forming a connection or waiting to receive a response) a RestClientException is raised. If
+     * the exception is due to timeout, the timeout is adjusted according to the HttpRequestTimeoutProgression object, and the request
+     * is reattempted. A final DropDeadInstant is also monitored and if it arrives a null response is returned.
+    */
+    public <T> ResponseEntity<T> exchange(String url, HttpMethod method, HttpEntity requestEntity, Map<String,?> uriVariables, Class<T> returnType) {
+		HttpRequestTimeoutProgression timeoutProgression = new HttpRequestTimeoutProgression(initialTimeout, maximumTimeout, dropDeadInstant);
```
n1zea144, line 115: Whitespace issue?

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpClientWithTimeoutAndRetry.java
```diff
@@ -0,0 +1,173 @@
...
+		SimpleClientHttpRequestFactory requestFactory = new SimpleClientHttpRequestFactory();
```
n1zea144, line 116: Whitespace issue?

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpClientWithTimeoutAndRetry.java
```diff
@@ -0,0 +1,173 @@
...
+        RestTemplate restTemplate = new RestTemplate();
+		restTemplate.setRequestFactory(requestFactory);
```
n1zea144, line 118: Whitespace issue?

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpClientWithTimeoutAndRetry.java
n1zea144, line 115: Whitespace issue - through line 118

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpClientWithTimeoutAndRetry.java
```diff
@@ -0,0 +1,173 @@
...
+                lastRestClientException = null; // reset before each request
+                lastResponseBodyStringAfterException = null; // reset before each request
+                numberOfRequestsAttempted = numberOfRequestsAttempted + 1;
+                lastTimeoutUsed = timeoutProgression.getNextTimeoutForRequest();
+		        requestFactory.setConnectTimeout(lastTimeoutUsed);
+		        requestFactory.setReadTimeout(lastTimeoutUsed);
+                ResponseEntity<T> response = restTemplate.exchange(url, method, requestEntity, returnType, uriVariables);
+                if (response.getStatusCode().is5xxServerError()) {
+                    pauseForMilliseconds(lastTimeoutUsed);
+System.out.println("got error: " + response.getStatusCode().toString());
```
n1zea144, line 138: Whitespace issue.

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpClientWithTimeoutAndRetry.java
n1zea144, line 139: Did you intend for this additional pause?

common/src/test/java/org/cbioportal/cmo/pipelines/common/util/UtilTestConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package org.cbioportal.cmo.pipelines.common.util;
+
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class UtilTestConfiguration {
+
```
n1zea144, line 39: Content?

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/SessionConfiguration.java
```diff
@@ -71,82 +73,105 @@ public class SessionConfiguration {
...
     * for CVR fetch for the portal, use 0
     */
-    private String dmpUrl;
-    private String gmlUrl;
 
-    // Gets the sessionId from CVR
+    /* Gets the sessionId from CVR
+     * TODO: this approach creates the session during the Bean Initialization phase (before actual application startup). It also
+     *       constrains each run of the cvr pipeline to having only a single session. Sessions expire after 3 hours, and this
+     *       does not allow for the possibility of begining subsequent sessions for later exchanges with the cvr servers if the
```
n1zea144, line 134: This means that we don't automatically renew a session and all work needs to be complete within 3 hrs.  Typo "beginning"

import-scripts/dmp-import-vars-functions.sh
```diff
@@ -319,35 +319,37 @@ function waitOutDmpGermlineServerInstabilityPeriod() {
 
 # Function for consuming fetched samples after successful import
 function consumeSamplesAfterSolidHemeImport {
+    drop_dead_instant_string=$(date --date="+2hours +50minutes" -Iseconds) # nearly 3 hours from now
```
n1zea144, line 322: Corresponds with cvr session timeout

import-scripts/fetch-dmp-data-for-import.sh
```diff
@@ -202,9 +202,9 @@ MY_FLOCK_FILEPATH="/data/portal-cron/cron-lock/fetch-dmp-data-for-import.lock"
 
     if [ $IMPORT_STATUS_IMPACT -eq 0 ] ; then
         # fetch new/updated IMPACT samples using CVR Web service   (must come after git fetching)
-        waitOutDmpTumorServerInstabilityPeriod
+        drop_dead_instant_string=$(date --date="+2hours +50minutes" -Iseconds) # nearly 3 hours from now for both tumor and germline fetching
```
n1zea144, line 205: Corresponds with cvr session timeout

import-scripts/fetch-dmp-data-for-import.sh
```diff
@@ -328,9 +325,9 @@ MY_FLOCK_FILEPATH="/data/portal-cron/cron-lock/fetch-dmp-data-for-import.lock"
 
     if [ $IMPORT_STATUS_HEME -eq 0 ] ; then
         # fetch new/updated heme samples using CVR Web service (must come after git fetching). Threshold is set to 50 since heme contains only 190 samples (07/12/2017)
-        waitOutDmpTumorServerInstabilityPeriod
+        drop_dead_instant_string=$(date --date="+2hours" -Iseconds) # 2 hours from now for both tumor and germline fetching
```
n1zea144, line 328: Why is this capped to 2 hours and not 3 like the others - same question for a few instances below.

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpClientWithTimeoutAndRetry.java
sheridancbio: literal tabs replaced with 4 space indent

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpClientWithTimeoutAndRetry.java
sheridancbio: fixed

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpClientWithTimeoutAndRetry.java
sheridancbio: This was testing code accidentally left in. Now removed.

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/InstantStringUtil.java
```diff
@@ -0,0 +1,274 @@
...
+import java.time.temporal.TemporalAccessor;
+import java.time.ZoneId;
+import java.time.zone.ZoneRulesException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.TimeZone;
+import org.apache.log4j.Logger;
+
+/* parses and converts strings to required format for java.time.Instant.parse(), which requires java.time.format.DateTimeFormatter.ISO_INSTANT compatible format.
+ * Several formats are attempted for parsing, based on easily output values from the unix 'date' program (run with LC_TIME="en_US.UTF-8"):
```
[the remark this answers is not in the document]
sheridancbio: Yes. The linux 'date' command can generate strings in number of different formats. If no format specifier is passed to 'date' the output format seems to be dependent on the locale settings configured on the system. This code could have been written to require one particular format (such as iso-8601 which has the advantage of not containing spaces) but users may not be familiar with reading/writing dates in that format. This code attempts to parse the date in several formats - those which were easily available in the java time library and which were easy to convert to from a "to the second" standard output format from 'date'. Should the default format of the date command change, or if a system migration changes the locale so that it is no longer compatible, having additional recognized formats allows easy adjustment without requiring java code changes.

common/src/test/java/org/cbioportal/cmo/pipelines/common/util/UtilTestConfiguration.java
sheridancbio: This class will be deleted if not needed. Any beans needed for unit testing (moving RestTemplateTimeoutTest into this package and adapting it to test HttpClientWithTimeoutAndRetry) - for now it is a placeholder.

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/SessionConfiguration.java
sheridancbio: Correct - that is the current behavior. One (single) session is obtained, and it expires in 3 hours (according to the comment in the response from lynx). Our code only attempts to obtain a session during the bean wiring phase of application startup. Improving this behavior so that session expiration does not prevent continued progress in a single run could be a future enhancement.

Typo corrected.

cvr/src/main/java/org/cbioportal/cmo/pipelines/CVRPipeline.java
```diff
@@ -182,43 +191,72 @@ private static void checkExceptions(JobExecution jobExecution, JobParameters job
...
+            }
+        }
+    }
+
+    public static void main(String[] args) throws Exception {
+        Options options = CVRPipeline.getOptions(args);
+        CommandLineParser parser = new DefaultParser();
+        CommandLine commandLine = parser.parse(options, args);
+        validateArguments(options, commandLine);
+        String dropDeadInstantString = "";
```
mandawilson, line 239: In this case `dropDeadInstantString` has a length of 0, and will be set to a distant time.

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/SessionConfiguration.java
```diff
@@ -1,15 +1,15 @@
 /*
- * Copyright (c) 2016 - 2017 Memorial Sloan-Kettering Cancer Center.
+ * Copyright (c) 2016, 2017, 2023 Memorial Sloan Kettering Cancer Center.
```
mandawilson, line 2: I assume we are not including years where the file wasn't modified.

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/consume/ConsumeSampleWriter.java
```diff
@@ -116,44 +127,53 @@ public void write(List<? extends String> sampleIdList) throws Exception {
...
+                dmpConsumeInitialResponseTimeout,
+                dmpConsumeMaximumResponseTimeout,
+                InstantStringUtil.createInstant(dropDeadInstantString),
+                false); // on a server error response, stop trying and fail/log (but continue on to other samples)
+        ResponseEntity<CVRConsumeSample> responseEntity = client.exchange(dmpConsumeUrl + sampleId, HttpMethod.GET, requestEntity, null, CVRConsumeSample.class);
+        if (responseEntity == null) {
+            String message = "";
+            if (client.getLastResponseBodyStringAfterException() != null) {
+                message = String.format("final response body was: '%s'", client.getLastResponseBodyStringAfterException());
+            } else {
```
mandawilson, line 146: `else if () {`

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/requeue/CvrRequeueTasklet.java
```diff
@@ -67,79 +68,93 @@ public class CvrRequeueTasklet implements Tasklet {
...
+                dmpRequeueSampleMaximumResponseTimeout,
+                InstantStringUtil.createInstant(dropDeadInstantString),
+                false); // on a server error response, stop trying and fail/log (but continue on to other samples)
+        String dmpRequeueUrl = String.format("%s%s/%s/%s", dmpServerName, dmpRequeue, sessionId, sampleId);
+        ResponseEntity<CVRRequeueRecord> responseEntity = client.exchange(dmpRequeueUrl, HttpMethod.GET, requestEntity, null, CVRRequeueRecord.class);
+        if (responseEntity == null) {
+            String message = "";
+            if (client.getLastResponseBodyStringAfterException() != null) {
+                message = String.format("final response body was: '%s'", client.getLastResponseBodyStringAfterException());
+            } else {
```
mandawilson, line 143: else if () {

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/samplelist/CvrSampleListsTasklet.java
```diff
@@ -172,50 +200,65 @@ private Set<String> loadWhitelistedSamplesWithZeroVariants() throws Exception {
...
+                dmpMasterListMaximumResponseTimeout,
+                InstantStringUtil.createInstant(dropDeadInstantString),
+                false); // on a server error response, stop trying and move on. We continue processing even when there is no retrieved master list
+        Set<String> dmpSamples = new HashSet<String>();
+        ResponseEntity<CVRMasterList> responseEntity = client.exchange(dmpUrl, HttpMethod.GET, requestEntity, null, CVRMasterList.class);
+        if (responseEntity == null) {
+            String message = "";
+            if (client.getLastResponseBodyStringAfterException() != null) {
+                message = String.format("final response body was: '%s'", client.getLastResponseBodyStringAfterException());
+            } else {
```
mandawilson, line 235: `else if () {`

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/variants/CVRVariantsProcessor.java
```diff
@@ -104,26 +110,40 @@ public String process(CvrResponse i) throws Exception {
...
+                dmpGetSegmentsMaximumResponseTimeout,
+                InstantStringUtil.createInstant(dropDeadInstantString),
+                false); // on a server error response, stop trying and move on. We accept samples even if they are missing their seg data
+        ResponseEntity<CVRSegData> responseEntity = client.exchange(dmpSegmentUrl, HttpMethod.GET, requestEntity, null, CVRSegData.class);
+        if (responseEntity == null) {
+            String message = "";
+            if (client.getLastResponseBodyStringAfterException() != null) {
+                message = String.format("final response body was: '%s'", client.getLastResponseBodyStringAfterException());
+            } else {
+                if (client.getLastRestClientException() != null) {
```
mandawilson, line 131: `else if () {`

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/variants/CvrResponseTasklet.java
```diff
@@ -32,61 +32,94 @@
...
+                dmpRetrieveVariantsInitialResponseTimeout,
+                dmpRetrieveVariantsMaximumResponseTimeout,
+                InstantStringUtil.createInstant(dropDeadInstantString),
+                true); // on a server error response, keep trying. If we cannot get the variants list, the overall fetch fails.
+        ResponseEntity<CvrResponse> responseEntity = client.exchange(dmpUrl, HttpMethod.GET, requestEntity, null, CvrResponse.class);
+        if (responseEntity == null) {
+            String message = "";
+            if (client.getLastResponseBodyStringAfterException() != null) {
+                message = String.format("final response body was: '%s'", client.getLastResponseBodyStringAfterException());
+            } else {
```
mandawilson, line 103: `else if () {`

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/variants/GMLVariantsReader.java
```diff
@@ -56,24 +58,46 @@ public class GMLVariantsReader implements ItemStreamReader<GMLVariant> {
...
+                dmpRetrieveVariantsMaximumResponseTimeout,
+                InstantStringUtil.createInstant(dropDeadInstantString),
+                true); // on a server error response, keep trying. If we cannot get the variants list, the overall fetch fails.
+        ResponseEntity<GMLVariant> responseEntity = client.exchange(dmpUrl, HttpMethod.GET, requestEntity, null, GMLVariant.class);
+        if (responseEntity == null) {
+            String message = "";
+            if (client.getLastResponseBodyStringAfterException() != null) {
+                message = String.format("final response body was: '%s'", client.getLastResponseBodyStringAfterException());
+            } else {
+                if (client.getLastRestClientException() != null) {
```
mandawilson, line 93: `else if () {`

import-scripts/dmp-import-vars-functions.sh
```diff
@@ -319,35 +319,37 @@ function waitOutDmpGermlineServerInstabilityPeriod() {
...
         echo "Consuming mskaccess samples from cvr"
-        $JAVA_BINARY $JAVA_CVR_FETCHER_ARGS -c $MSK_ACCESS_PRIVATE_DATA_HOME/cvr_data.json
+        $JAVA_BINARY $JAVA_CVR_FETCHER_ARGS -c $MSK_ACCESS_PRIVATE_DATA_HOME/cvr_data.json -z $drop_dead_instant_string
         rm -f $MSK_ACCESS_CONSUME_TRIGGER
     fi
 }
 
 # Function for consuming fetched samples after successful archer import
 function consumeSamplesAfterArcherImport {
+    drop_dead_instant_string=$(date --date="+1hour" -Iseconds) # one hour from now
```
mandawilson, line 348: Ben asked below about some timeouts that are 2 hours instead of nearly 3.  This one is 1 hour.  Maybe you explain somewhere.

import-scripts/fetch-dmp-data-for-import.sh
sheridancbio: The drop dead instant should be set to the point in the future where we want to abandon the current attempt to fetch samples from the cvr sample.

Looking at heme fetches at various points in the last month:
(tumor start)-->(tumor end) /// (germline start)-->(germline end)
23:11:04 --> 1:27:55 /// 2:30:09 --> 02:31:06
23:30:33 --> 1:49:40 /// 2:3023 --> 02:31:03
23:29:52 --> 2:10:41 /// 2:30:28 --> 2:31:09
23:36:05 --> 2:10:25 /// 2:30:09 --> 2:30:50
23:27:33 --> 2:23:12 /// 2:30:09 --> 2:30:51  (This was august 31 -> september 1)

The gap of time between the two fetches is for the current script code pause during the call to waitOutDmpTumorServerInstabilityPeriod

This shows that I was incorrect thinking that the heme fetch completed relatively quickly each night .. so thank you for pointing this out. The truth is quite the opposite --- the tumor sample fetches above all took more than 2 hours, and the last on was just short of 3 hours. This shows an urgency for us to address the single session per job execution problem with a maximum runtime of 3 hours.

It may not be as bad as it seems ... during a standard cvr job the order of operation is basically:
- get session id
- download the master list for the study
- fetch the json of all queued samples
- for each sample fetched in the previous step, fetch segment data for it (CVRVariantsProcessor)
- do all the processing of datatypes ... including annotation of variants
- requeue samples on the master list but which are not in our datafiles

So -- we basically have all the data before we start the mutation annotation step, which is the longest running step usually. For example, for the nearly 3 hour run on August 31, the mutation annotation started at 23:55:16 .. which is 28 minutes after startup. We will not need the session token again until the final (requeue) step, so if the token expires we will fail to requeue that night (maybe we will succeed the next night).

I suggest that we do prioritize the renewal of the session token when it is nearing expiration. But for the purposes of this PR I will make the drop_dead_instant be 3 hours after the start of the heme tumor fetch, and also I will have a separate drop_dead_instant for the germline fetch (perhaps giving that a 1 hour drop dead limit). Overall, we want to make sure that the dmp pipeline will fail and exit if the dmp server is not responsive at all. So I think I actually need to put some logic in the script to have a timestamp for the total process (of all cohorts) to complete. We do about 6 fetches, so if each waits for 3 hours to fail the whole process to final failure may take 18 hours, which is too much.

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpClientWithTimeoutAndRetry.java
```diff
@@ -0,0 +1,171 @@
...
+        this.numberOfRequestsAttempted = 0;
+        this.lastTimeoutUsed = 0;
+        this.reachedDropDeadInstant = false;
+        this.lastRestClientException = null;
+    }
+
+    public HttpClientWithTimeoutAndRetry(int initialTimeout, int maximumTimeout, Instant dropDeadInstant, boolean retryOnErroneousServerResponse) {
+        this.initialTimeout = initialTimeout;
+        this.maximumTimeout = maximumTimeout;
+        this.dropDeadInstant = dropDeadInstant;
```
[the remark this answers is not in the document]
sheridancbio: fixed. Now dropDeadInstant should never be null. It will be set to a default value if null is pased in.

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpClientWithTimeoutAndRetry.java
```diff
@@ -0,0 +1,171 @@
...
+        restTemplate.setRequestFactory(requestFactory);
+        if (uriVariables == null) {
+            uriVariables = new HashMap<String, String>();
+        }
+        numberOfRequestsAttempted = 0;
+        lastTimeoutUsed = 0;
+        reachedDropDeadInstant = false;
+        lastRestClientException = null;
+        lastResponseBodyStringAfterException = null;
+        while (Instant.now().isBefore(dropDeadInstant)) {
```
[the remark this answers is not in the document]
sheridancbio: good catch. Fixed.

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpClientWithTimeoutAndRetry.java
```diff
@@ -0,0 +1,171 @@
...
+    public RestClientException getLastRestClientException() {
+        return lastRestClientException;
+    }
+
+    public String getLastResponseBodyStringAfterException() {
+        return lastResponseBodyStringAfterException;
+    }
+
+    public boolean exceptionCausedByTimeout(RestClientException e) {
+        return e.getRootCause() instanceof SocketTimeoutException && e.getRootCause().getMessage().toLowerCase().contains("timed out");
```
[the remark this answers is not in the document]
sheridancbio: I do not know the answer, but I did see that the message included the text "timed out" in the timeouts caused by the exceeding of what was specified in setConnectTimeout() and setReadTimeout(). Those are the cases where I wanted our system to continue looping/retrying without logging. I kind of doubt that there are other SocketTimeoutException objects (which do not say "timed out" in the message), but just in case there are, these would be a little unexpected, so they would in fact be logged (in logger.debug()). Either way, the retry loop continues. I guess the difference is that expected timeout exceptions are not logged.

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpClientWithTimeoutAndRetry.java
```diff
@@ -0,0 +1,171 @@
...
+                if (!retryOnErroneousServerResponse) {
+                    return null; // fail now : the exception and response body have been captured in instance variables
+                }
+                pauseForMilliseconds(lastTimeoutUsed);
+            } catch (RestClientException e) {
+                // these exceptions are either timeouts or other low level exceptions. Continue to attempt the request.
+                lastRestClientException = e;
+                if (!exceptionCausedByTimeout(e)) {
+                    // some other low level issue (IOException usually - maybe NoRouteToHost if interface is down?)
+                    log.debug(String.format("RestClientException ocurred during retry loop : %s", e.toString()));
```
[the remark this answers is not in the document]
sheridancbio: Correct. I suppose we could log the RestClientResponseExceptions too. However we capture the RestClientResponseExceptions and make them available through getLastRestClientException() and getLastResponseBodyStringAfterException(). These methods are called in the appropriate places to do logging similar to the logging which had previously been done. (Different clients locations had different logging approaches)

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpRequestTimeoutProgression.java
```diff
@@ -0,0 +1,90 @@
...
+
+    private int initialTimeout; // the first configured timeout to be used before retry (in milliseconds)
+    private int maximumTimeout; // the upper limit for subsequent configured timeouts before retry (in milliseconds)
+    private Instant dropDeadMoment; // a fixed time at which any current attempt and any further attempt will be abondoned, returning with failure
+    private int timeoutForNextQuery;
+
+    public HttpRequestTimeoutProgression() {
+        initialTimeout = 1000;
+        maximumTimeout = 60000;
+        dropDeadMoment = Instant.now().plusSeconds(4 * maximumTimeout); // default alows 19 timeout loops to complete, ranging from 1 second to 60 seconds (the last three iterations have 60 second timeouts.)
```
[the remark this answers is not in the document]
sheridancbio: Thank you for pointing at this. The call to .plusSeconds() should have been to .plusMillis() I will fix that. I will also call the method for making the defaultDropDeadInstant which I already added for HttpClientWithTimeoutAndRetry

common/src/test/java/org/cbioportal/cmo/pipelines/common/util/UtilTestConfiguration.java
sheridancbio: removed

import-scripts/fetch-dmp-data-for-import.sh
sheridancbio: The overall runtime for script fetch-dmp-impact-data is currently 11 hours (starting at 7pm and ending around 5:45am). I have imposed an overall limit of 12 hours for this script to complete. We also have logic in fetch-and-import-dmp-data-wrapper which skips over the import steps if it is after 7am anyway. The consume samples step later (during the import step) has been set to a 3 hour timeout. In the worst possible case, the dmp fetching and dmp importing is constrained to complete in 18 hours, so multi-day stalls should be avoided if the dmp server returns to normal service.

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpRequestTimeoutProgression.java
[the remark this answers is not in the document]
sheridancbio: added and fixed

common/src/main/java/org/cbioportal/cmo/pipelines/common/util/HttpRequestTimeoutProgression.java
```diff
@@ -0,0 +1,90 @@
...
+        timeoutForNextQuery = initialTimeout;
+    }
+
+/* if dropDeadMoment is null, by default the dropDeadMoment will be set to a point in the future 4 times the maximum timeout from now
```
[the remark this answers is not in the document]
sheridancbio: fixed

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/SessionConfiguration.java
sheridancbio: Basically yes, although I did not always check the git history to see which years modifications actually happened in. We should probably switch to using a project-wide copyright header which is identical across all source files. If any modification to any file occurs within a year, then all headers across the project will be updated even for individual files which did not change. Authors revise books, and they do not track changes to individual chapters (or pages). Possibly we could have a uniform copyright header per-module or other distributable artifact.

cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/consume/ConsumeSampleWriter.java
sheridancbio: following google java format, the syntactically optional braces for blocks after "if" "else" "for" "do" and "while" are always to be used:

https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

import-scripts/dmp-import-vars-functions.sh
sheridancbio: explanation and adjustments added above in Benjamin's comment. Timeouts have been largely standardized to 3 hours per fetcher execution now, with an end-to-end timeout for the entire fetch script set to 12 hours.

sheridancbio:
Regression testing performed for mskimpact cohort, comparing branch head to changeset 5f2bf4b837f601b3ca43137e188f70f3d5d429ce (Fri Jul 28 13:51:33 2023) . Aside from column re-ordering and row re-ordering, the following datatypes were seen to match:
- data_mutation_extended.txt
- zero_variant_whitelist.txt
- seq_date.txt
- data_nonsignedout_mutations.txt
- data_CNA.txt
- mskimpact_data_cna_hg19.seq
- data_gene_matrix.txt
- data_sv.txt
- data_clinical_sample.txt (exported from redcap)
- data_clinical_patient.txt (exported from redcap)
- data_timeline.txt (exported from redcap)
To speed testing, the starting size of the mskimpact data_nonsignedout_mutations.txt and data_CNA.txt and mskimpact_data_cna_hg19.seq were reduced to allow faster reading. Modification to the crdb fetcher and ddp fetcher during the update to java 11 were minimal and regression testing was not performed. Additionally, we a no longer running the CRDB fetcher as part of our nightly pipeline.

Regression testing was attempted on the darwin fetcher, but although the code compiled without error the darwin pipeline fails to run under java 11. Compiling under java 8 and then running under java 11 also failed (both with initialization errors). Errors were similar to the errors which required an update to the mockito version, however updating querydsl libraries failed to produce a runnable build. For now it seems that the darwin fetcher is broken .. and was already broken at the changeset 5f2bf4b837f601b3ca43137e188f70f3d5d429ce before the updates for java11. Short of debugging and updating our darwin fetcher code to use a modern version of JPA/mysql/querydsl, I think we must live with a frozen darwin fetcher artifact for now. We plan to move away from direct querying of the darwin caisis databases soon in any case.

sheridancbio:
import/fetch scripts updated to use the appropriate java version for importers/fetchers. CVR and REDCAP_PIPELINE runs now operate to JAVA 11 standards (under the openjdk-19 interpreter)

sheridancbio:
Code merged without waiting for [fetcher-redcap-schema-integration-test](http://dashi-dev.cbio.mskcc.org:8090/job/fetcher-redcap-schema-integration-test/) to complete. It had succeeded on previous versions which contained all java code updates, and these tests take 40 minutes to run. Also - our jenkins server should be updated to include the java19 version from openjdk and jvm selection based on which artifact is needed to be built or tested.

PR #1071, opened 2023-09-16
cvr seg data fetch retry fix

sheridancbio:
- consider a response from the server which does not fit the model a server error. Seg data fetches are not reattempted on server errors, so this will prevent retry when the server has reported that seg data cannot be fetched for a sample. (in a 200 OK response)


sheridancbio:
Merging without waiting for fetcher-redcap-schema-integration-test. Test takes 40 minutes and no changes to schema were made.

PR #1074, opened 2023-09-20
Fix bug that is preventing cna seg file from updating for az_mskimpact

callachennault:
The `az_mskimpact_data_cna_hg19.seg` file is not being updated in the az_mskimpact cohort because of a bug in the `update-az-mskimpact.sh` script. The script copies over all files from msk_solid_heme and filters out files that we do not want to deliver. The CNA seg file is named `mskimpact_data_cna_hg19.seg` for msk_solid_heme, and the script was not renaming the updated file to `az_mskimpact_data_cna_hg19.seg`, so the file containing updated data was being deleted on each run and not committed. This PR fixes that bug.

### DIFF
--- a/import-scripts/update-az-mskimpact.sh
+++ b/import-scripts/update-az-mskimpact.sh
@@ -107,6 +107,29 @@ function filter_files_in_delivery_directory() {
     return 0
 }
 
+function rename_files_in_delivery_directory() {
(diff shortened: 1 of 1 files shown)


averyniceday (approved):
lgtm, good catch!

PR #1083, opened 2023-10-16
Migrate cmo-pipelines to Java11 

averyniceday:
- Tested on dev machine - build and ran using java11 
- Updates to POM for moved/dropped depencies
- Updates to DDP pipeline for deprecated libs


import-scripts/knowledgesystems-importer/automation-environment.sh
```diff
@@ -52,16 +52,7 @@ export_java_home "AUTODETECT"
...
-export JAVA_8_BINARY="$JAVA_8_HOME/jre/bin/java"
-export JAVA_19_HOME="/usr/lib/jdk-19.0.2"
-export JAVA_19_BINARY="$JAVA_19_HOME/bin/java"
-export JAVA_BINARY_FOR_CRDB="$JAVA_8_BINARY"
-export JAVA_BINARY_FOR_CVR="$JAVA_19_BINARY"
-export JAVA_BINARY_FOR_DARWIN="$JAVA_8_BINARY"
-export JAVA_BINARY_FOR_DDP="$JAVA_8_BINARY"
-export JAVA_BINARY_FOR_REDCAP="$JAVA_19_BINARY"
-export JAVA_BINARY_FOR_IMPORTER="$JAVA_8_BINARY"
+export JAVA_HOME="/usr/lib/jvm/java-11-openjdk-11.0.20.0.8-1.amzn2.0.1.x86_64"
```
sheridancbio, line 55: wouldn't this need to be moved above line 54? (which depends on JAVA_HOME?)
The functions above this detect JAVA_HOME based on the location of java with the command "which java". Another way to manage this is to delete line 55 and adjust /home/cbioportal_importer/.bashrc so that the path to the java 11 jvm comes earlier in the PATH environment than any other JVM path. Note that this should be tested for both interactive shells and non-interactive shells (such as those launched by crontab).

import-scripts/pipelines_eks/automation-environment.sh
```diff
@@ -52,16 +52,7 @@ export_java_home "AUTODETECT"
...
-export JAVA_8_BINARY="$JAVA_8_HOME/jre/bin/java"
-export JAVA_19_HOME="/usr/lib/jdk-19.0.2"
-export JAVA_19_BINARY="$JAVA_19_HOME/bin/java"
-export JAVA_BINARY_FOR_CRDB="$JAVA_8_BINARY"
-export JAVA_BINARY_FOR_CVR="$JAVA_19_BINARY"
-export JAVA_BINARY_FOR_DARWIN="$JAVA_8_BINARY"
-export JAVA_BINARY_FOR_DDP="$JAVA_8_BINARY"
-export JAVA_BINARY_FOR_REDCAP="$JAVA_19_BINARY"
-export JAVA_BINARY_FOR_IMPORTER="$JAVA_8_BINARY"
+export JAVA_HOME ="/usr/lib/jvm/java-11-openjdk-11.0.20.0.8-1.amzn2.0.1.x86_64"
```
sheridancbio, line 55: see previous comment

sheridancbio (commented):
This looks good ... unifying to a single java installation for all java executions. However, see comments below about the proper setting of JAVA_HOME and JAVA_BINARY so that the correct JVM is used in both interactive and non-interactive shells.

Ah .. I see what we are doing (I think). The PATH resolves 'java' to /etc/alternatives/java on our server, which is a symlink to /usr/lib/jvm/java-11-openjdk-11.0.20.0.8-1.amzn2.0.1.x86_64/bin/java  ... so ok, the functions in automation_environment which locate JAVA_PATH will succeed in setting JAVA_PATH appropriately. This means that the extra line in automation_environment explicitly setting JAVA_PATH is not needed (but not harmful)

I suggest also setting the other "java-related" executable symlinks to the new home path (javac javap javah javadoc jstack jdb ...) otherwise we may find ourselves doing maven builds with java11 but trying to debug / stacktrace with java8.

sheridancbio (approved):

PR #1086, opened 2023-10-25
CDM Integration

averyniceday:
- update merge script to optionally merge specific file types
- add new vars for CDM paths
- add section for git checkout, merge, copy of CDM data


mandawilson (changes requested):
Looks good, just check the part where we still copy files into the repo even if the merge has failed or we fail to add the headers.

mandawilson (approved):
Great!

PR #1093, opened 2023-11-22
Fix syntax bug in mskimpact redcap export

callachennault:
Possible syntax bug in list of projects to ignore when exporting mskimpact from redcap


callachennault:
No longer needed - ignored projects were removed from redcap export call

PR #1094, opened 2023-11-27
remove references to DDP/CRDB/Darwin

averyniceday:


averyniceday:
Rob and Manda looked at this and approved

PR #819, opened 2021-02-03
GENIE subset fixes

rmadupuri:
Fixes made:
1. **CNA Gene Merge Tool:** https://github.com/knowledgesystems/cmo-pipelines/issues/653
Script that will take a CNA file and perform the following gene merges:
CDKN2Ap16/14 -> CDKN2A
MLLs -> KMTs (MLL, MLL2, MLL3 and MLL4 -> KMT2A KMT2B KMT2C KMT2D)
FAM123B -> AMER1
MYCL1 -> MYCL

2. **Remove duplicate variants from MAF**
Script to remove duplicate maf records based on the 8 key columns (Entrez_Gene_Id, Chromosome, Start_Position, End_Position, Variant_Classification, Tumor_Seq_Allele2, Tumor_Sample_Barcode, HGVSp_Short)
Calculates VAF for each record and picks the record with high VAF
Formula for VAF = t_alt_count / (t_ref_count + t_alt_count) 

3. **Missing patient ids in patient file** https://github.com/knowledgesystems/cmo-pipelines/issues/654
Patients that have no attributes only appear in the sample clinical data file.

4. **Vital status location change** https://github.com/knowledgesystems/cmo-pipelines/issues/657
Patient vital status attributes were moved from the patient file into vital_status.txt. Since they GENIE has reversed that decision and now wants vital status information back in the patient clinical file.

5. **Change value for clinical attributes with no value** https://github.com/knowledgesystems/cmo-pipelines/issues/656
AGE_AT_SEQ_REPORT, ONCOTREE_CODE, BIRTH_YEAR, YEAR_DEATH, YEAR_CONTACT, INT_CONTACT, INT_DOD cannot be NA or blank. All attribute values should be "Unknown" when the patient/attribute or sample/attribute pair is missing a value.
Also, for attributes that take on NAACR values, like SEX, PRIMARY_RACE, SECONDARY_RACE, TERTIARY_RACE, ETHNICITY, they should take on the NAACR value for NA - (SEX = 9, RACE = 99, ETHNICITY = 9)

6. **Gene panel names missing prefix in clinical file** https://github.com/knowledgesystems/cmo-pipelines/issues/655
In the "Sequence Assay ID" column of the data_clinical_sample.txt file, IMPACT panel names should be prefixed with an "MSK-".

PR #1099, opened 2023-12-06
Fix bug in checking for duplicate Mutation Records

callachennault:
- This PR corrects the `isDuplicateRecord()` check in `CVRMutationDataReader`, `CVRNonSignedoutMutationDataReader`, and `GMLMutationDataReader`. Duplicate records were not being caught by this check because the map being used to check for duplicates was not being populated. The `addRecordToMap()` function now adds to the `mutationMap` and the duplicate check functions correctly.
- This PR also adds the `remove-duplicate-maf-variants.py` script provided by the curation team. Functionality of script described [here](https://github.com/knowledgesystems/cmo-pipelines/pull/819). It is now called in the `update-az-mskimpact.sh` script

### DIFF
--- a/cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/mutation/CVRMutationDataReader.java
+++ b/cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/mutation/CVRMutationDataReader.java
@@ -173,9 +173,6 @@ public class CVRMutationDataReader implements ItemStreamReader<AnnotatedRecord> 
         MutationRecord to_add;
         while ((to_add = reader.read()) != null && to_add.getTUMOR_SAMPLE_BARCODE() != null) {
             // skip if new sample or if mutation record for sample seen already
-            String tumorSampleBarcode = to_add.getTUMOR_SAMPLE_BARCODE();
-            String hugoSymbol = to_add.getHUGO_SYMBOL();
-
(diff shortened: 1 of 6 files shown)


cvr/src/main/java/org/cbioportal/cmo/pipelines/cvr/mutation/CVRMutationDataReader.java
```diff
@@ -132,7 +132,9 @@ private void loadMutationRecordsFromJson(CVRData cvrData) {
             String somaticStatus = result.getMetaData().getSomaticStatus() != null ? result.getMetaData().getSomaticStatus() : "N/A";
             int countSignedOutSnps = result.getAllSignedoutCvrSnps().size();
             for (CVRSnp snp : result.getAllSignedoutCvrSnps()) {
-                recordsToAnnotate.add(cvrUtilities.buildCVRMutationRecord(snp, sampleId, somaticStatus));
+                MutationRecord to_add = cvrUtilities.buildCVRMutationRecord(snp, sampleId, somaticStatus);
+                recordsToAnnotate.add(to_add);
+                addRecordToMap(to_add);
```
averyniceday, line 137: Think this logic will fix the issue, but have you run an entire mskimpact fetch with this code? I'm curious about the memory usage, worried that if we have another map tracking the entire mutation record, we'll exacerbate the memory issue

PR #1105, opened 2024-01-17
Add sophia script

callachennault:
- [x] WIP: Need to complete testing - make sure this produces same results as previous script

Adds Sophia cohort generation script to import-scripts directory. Note that this script and the update-az-mskimpact.sh script have some minor overlap (report_error, rename_files_in_delivery_directory, add_metadata_headers) and should be generalized.


import-scripts/update-az-mskimpact.sh
```diff
@@ -62,7 +61,7 @@ function transfer_to_az_sftp_server() {
     # Connect and transfer data
     # With use of here-doc, there must be no leading whitespace until EOF
     sftp -i "$TRANSFER_KEY" "$SFTP_USER"@"$SERVICE_ENDPOINT" -b <<EOF
-put -R "$AZ_DATA_HOME/az_mskimpact" "$AZ_REPO_NAME"
+put -R "$AZ_DATA_HOME/$AZ_MSKIMPACT_STABLE_ID" "$AZ_REPO_NAME"
```
averyniceday, line 64: you have `AZ_MSK_IMPACT_DATA_HOME` defined in automation_environment

import-scripts/pipelines_eks/automation-environment.sh
```diff
@@ -188,6 +189,7 @@ export LYMPHOMA_SUPER_COHORT_DATA_HOME=$DMP_DATA_HOME/lymphoma_super_cohort_fmi_
 export MSK_EXTRACT_COHORT_DATA_HOME=$MSK_MIND_DATA_HOME/datahub/msk_extract_cohort2_2019
 export MSK_SPECTRUM_COHORT_DATA_HOME=$MSK_SHAHLAB_DATA_HOME/msk_spectrum
 export MSK_CHORD_DATA_HOME=$CDSI_DATA_HOME/msk-chord
+export AZ_MSK_IMPACT_DATA_HOME="$AZ_DATA_HOME/az_mskimpact"
```
averyniceday, line 192: add sophia home here

import-scripts/transpose_cna_py3.py
```diff
@@ -0,0 +1,25 @@
+import argparse
```
averyniceday, line 1: let's add some comments here about overall usage of the script - I know it's in the description but maybe on the top about use case and restrictions on passed in files

also can we add some comments about pandas functions -- think as a whole we're not too familiar so hard to read the code. `eg. df.T`

import-scripts/update-sophia-mskimpact.sh
```diff
@@ -0,0 +1,256 @@
+#!/usr/bin/env bash
+
+# File containing list of patients should be passed in as argument
+export SUBSET_FILE="$1"
+export CANCER_TYPE="$2"
+export CURRENT_DATE="$(date '+%m.%d.%y')"
+export COHORT_NAME="sophia-$CANCER_TYPE-data-$CURRENT_DATE"
+
+export SOPHIA_REPO_NAME="sophia-data"
+export SOPHIA_DATA_HOME="$PORTAL_DATA_HOME/$SOPHIA_REPO_NAME"
```
averyniceday, line 10: switch to the automation environment method of setting vars

import-scripts/update-az-mskimpact.sh
callachennault: good catch, just fixed

import-scripts/update-sophia-mskimpact.sh
```diff
@@ -0,0 +1,256 @@
...
+export SOPHIA_MSKIMPACT_STABLE_ID="sophia_mskimpact"
+export SOPHIA_COHORT_HOME="$SOPHIA_DATA_HOME/deliveries/$COHORT_NAME"
+export SOPHIA_MSK_IMPACT_DATA_HOME="$SOPHIA_COHORT_HOME/$SOPHIA_MSKIMPACT_STABLE_ID"
+export SOPHIA_TMPDIR="$SOPHIA_COHORT_HOME/tmp"
```
averyniceday, line 14: so if I understand right -- 

sophia_portal_data/deliveries/<cancer-type>-<datespecific>/sophia_mskimpact

Is there anything parallel to deliveries? 

Also for tmp function, let's use something on the tmp volume - think we autogenerate a tmp directory as part of the script. If that's here, wil we still need the `sophia_mskimpact` directory?

averyniceday: Was thinking something closer to 

cbio-portal-data/delivery-pipelines/<recepient e.g. sophia>/<cancer-type-date>

callachennault: Right now `/data/portal-cron/cbio-portal-data/sophia-data` contains these directories/files:
`cohort_ids/`: contains files with list of IDs for each cohort
`deliveries/`: contains directories for each cancer type cohort and a date (in the case that we have to make changes / deliver multiple versions)
- for example, a subdirectory would be `sophia-lung-data-01.28.24` and that directory contains: `README.pdf`, `gene_panels/` directory, and `sophia-mskimpact/` directory containing study files. This structure was copied from the structure of the az repo but since sophia doesn't import these files to a cbio instance it's probably not totally necessary, just thought it was a good idea to keep the same structure
`gene_panels/`: gene panel files that are packaged into each delivery
`logs/`: log files from cohort generation
`README.pdf`: readme file describing files that is packaged into each delivery

if you want to talk through this setup @averyniceday let me know, totally open to opinions

import-scripts/pipelines_eks/automation-environment.sh
callachennault: done

import-scripts/transpose_cna_py3.py
callachennault: done

averyniceday (approved):

PR #1107, opened 2024-02-01
Modify preconsume script to work on one cohort at a time

callachennault:
This PR :
- modifies the `preconsume_problematic_samples.sh` script to work on one cohort at a time. 
- Modifies the `fetch-dmp-data-for-import.sh` script to call the `preconsume_problematic_samples.sh` once for each cohort, immediately before the CVR fetch for that cohort.
- This should prevent problematic samples from being queued in the time between the preconsume step and the CVR fetcher step.

### DIFF
--- a/import-scripts/fetch-dmp-data-for-import.sh
+++ b/import-scripts/fetch-dmp-data-for-import.sh
@@ -185,10 +185,8 @@ MY_FLOCK_FILEPATH="/data/portal-cron/cron-lock/fetch-dmp-data-for-import.lock"
         # fetch new/updated IMPACT samples using CVR Web service   (must come after git fetching)
         drop_dead_instant_step=$(date --date="+3hours" -Iseconds) # nearly 3 hours from now
         drop_dead_instant_string=$(find_earlier_instant "$drop_dead_instant_step" "$DROP_DEAD_INSTANT_END_TO_END")
-        
         # pre-consume any samples missing values for normal_ad/tumor_dp/ad fields, or having UNKNOWN gene-panel
         bash $PORTAL_HOME/scripts/preconsume_problematic_samples.sh mskimpact
-
         printTimeStampedDataProcessingStepMessage "CVR fetch for mskimpact"
         $JAVA_BINARY $JAVA_CVR_FETCHER_ARGS -d $MSK_IMPACT_DATA_HOME -p $MSK_IMPACT_PRIVATE_DATA_HOME -n data_clinical_mskimpact_data_clinical_cvr.txt -i mskimpact -r 150 $CVR_TEST_MODE_ARGS -z $drop_dead_instant_string
         if [ $? -gt 0 ] ; then
@@ -266,10 +264,8 @@ MY_FLOCK_FILEPATH="/data/portal-cron/cron-lock/fetch-dmp-data-for-import.lock"
         # fetch new/updated heme samples using CVR Web service (must come after git fetching). Threshold is set to 50 since heme contains only 190 samples (07/12/2017)
         drop_dead_instant_step=$(date --date="+3hours" -Iseconds) # nearly 3 hours from now
         drop_dead_instant_string=$(find_earlier_instant "$drop_dead_instant_step" "$DROP_DEAD_INSTANT_END_TO_END")
-        
         # pre-consume any samples missing values for normal_ad/tumor_dp/ad fields, or having UNKNOWN gene-panel
         bash $PORTAL_HOME/scripts/preconsume_problematic_samples.sh mskimpact_heme
-
         printTimeStampedDataProcessingStepMessage "CVR fetch for hemepact"
         $JAVA_BINARY $JAVA_CVR_FETCHER_ARGS -d $MSK_HEMEPACT_DATA_HOME -p $MSK_HEMEPACT_PRIVATE_DATA_HOME -n data_clinical_hemepact_data_clinical.txt -i mskimpact_heme -r 50 $CVR_TEST_MODE_ARGS -z $drop_dead_instant_string
         if [ $? -gt 0 ] ; then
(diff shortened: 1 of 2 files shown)


import-scripts/preconsume_problematic_samples.sh
```diff
@@ -124,14 +125,18 @@ function consume_hardcoded_samples() {
     rm -f ${PROBLEMATIC_EVENT_CONSUME_IDS_FILEPATH} ${PROBLEMATIC_METADATA_CONSUME_IDS_FILEPATH}
     touch ${PROBLEMATIC_EVENT_CONSUME_IDS_FILEPATH}
     touch ${PROBLEMATIC_METADATA_CONSUME_IDS_FILEPATH}
-    echo "P-0025907-N01-IM6" >> "${PROBLEMATIC_METADATA_CONSUME_IDS_FILEPATH}"
```
averyniceday, line 127: I mean, know this was here before, but any idea what's this for 😆 Is it not getting caught by our usual script

callachennault: yeah I looked into it, this is a sample we're receiving from the tumor server but it has a normal identifier. Rob added it to this `consume_hardcoded_samples` function until we got more info/a fix from CVR. I reached out to Mihir about it again and he said to continue consuming these and he'll get back to me if there's any change to it

import-scripts/preconsume_problematic_samples.sh
```diff
@@ -36,27 +32,32 @@ function make_tmp_dir_if_necessary() {
...
+    if [ "$COHORT" == "mskimpact" ] ; then
+        CVR_FETCH_URL_PREFIX=$CVR_IMPACT_FETCH_URL_PREFIX
+    elif [ "$COHORT" == "mskimpact_heme" ] ; then
+        CVR_FETCH_URL_PREFIX=$CVR_HEME_FETCH_URL_PREFIX
+    elif [ "$COHORT" == "mskarcher" ] ; then
+        CVR_FETCH_URL_PREFIX=$CVR_ARCHER_FETCH_URL_PREFIX
+    elif [ "$COHORT" == "mskaccess" ] ; then
+        CVR_FETCH_URL_PREFIX=$CVR_ACCESS_FETCH_URL_PREFIX
+    else
+        echo "Error: Invalid cohort name '$COHORT'" >&2
```
averyniceday, line 45: not a big deal, but let's move this check out into it's own function/block

averyniceday (approved):
small suggestion for change, but otherwise looks good!

import-scripts/fetch-dmp-data-for-import.sh
```diff
@@ -188,6 +185,10 @@ MY_FLOCK_FILEPATH="/data/portal-cron/cron-lock/fetch-dmp-data-for-import.lock"
         # fetch new/updated IMPACT samples using CVR Web service   (must come after git fetching)
         drop_dead_instant_step=$(date --date="+3hours" -Iseconds) # nearly 3 hours from now
         drop_dead_instant_string=$(find_earlier_instant "$drop_dead_instant_step" "$DROP_DEAD_INSTANT_END_TO_END")
+
```
sheridancbio, line 188: I think the comment before the function call is enough of a separator. I think I'd prefer to remove the blank line before and after the call to the preconsume script.

import-scripts/preconsume_problematic_samples.sh
```diff
@@ -36,27 +32,32 @@ function make_tmp_dir_if_necessary() {
...
 
 function detect_samples_with_problematic_events() {
-    $DETECT_SAMPLES_WITH_NULL_DP_AD_FIELDS_SCRIPT_FILEPATH ${IMPACT_FETCH_OUTPUT_FILEPATH} ${IMPACT_CONSUME_IDS_FILEPATH}
-    $DETECT_SAMPLES_WITH_NULL_DP_AD_FIELDS_SCRIPT_FILEPATH ${HEME_FETCH_OUTPUT_FILEPATH} ${HEME_CONSUME_IDS_FILEPATH}
-    $DETECT_SAMPLES_WITH_NULL_DP_AD_FIELDS_SCRIPT_FILEPATH ${ACCESS_FETCH_OUTPUT_FILEPATH} ${ACCESS_CONSUME_IDS_FILEPATH}
+    $DETECT_SAMPLES_WITH_NULL_DP_AD_FIELDS_SCRIPT_FILEPATH ${FETCH_OUTPUT_FILEPATH} ${PROBLEMATIC_EVENT_CONSUME_IDS_FILEPATH}
 }
 
 function detect_samples_with_problematic_metadata() {
-    $DETECT_SAMPLES_WITH_PROBLEMATIC_METADATA_SCRIPT_FILEPATH ${ARCHER_FETCH_OUTPUT_FILEPATH} ${ARCHER_CONSUME_IDS_FILEPATH}
```
sheridancbio, line 54: I guess before we were only scanning archer for genepanel=UNKNOWN. This change seems to mean that we will scan all 4 cohorts for bad genepanel references (probably a good thing). Does this run smoothly though on other cohorts? (The json schema may differ)

import-scripts/preconsume_problematic_samples.sh
```diff
@@ -140,11 +145,11 @@ function log_actions() {
...
     if [ ${#failed_to_consume_problematic_events_sample_list[@]} -gt 0 ]; then
         MESSAGE="${MESSAGE} Attempted Unsuccessfully To Consume :\n${failed_to_consume_problematic_events_sample_list[*]}"
     fi
-    MESSAGE="${MESSAGE}Warning : the following samples have been reemptively consumed before fetch because they contained problematic metadata where the gene-panel property was unset or had value UNKNOWN.\nSuccessfully Consumed :\n${succeeded_to_consume_problematic_metadata_sample_list[*]}"
+    MESSAGE="${MESSAGE}Warning : the following samples have been preemptively consumed before fetch because they contained problematic metadata where the gene-panel property was unset or had value UNKNOWN.\nSuccessfully Consumed :\n${succeeded_to_consume_problematic_metadata_sample_list[*]}"
```
sheridancbio, line 152: nice typo fix

sheridancbio (approved):
Nice work. My only question is whether the script detect_samples_with_problematic_metadata.py will run correctly on the impact / impact-heme / access cohorts. If we don't want to worry about whether it runs correctly or not we could skip over the scan for metadata problems in these datasets. But the metadata is at the top level of the json object returned for samples, so it probably is compatible to all cohorts.

import-scripts/preconsume_problematic_samples.sh
callachennault: Yes, with the way I rewrote it, all cohorts will be checked for problematic events + problematic metadata. I figured it made sense since we already have the functionality. I didn't run into any errors testing on the 4 cohorts but the queues that I pulled didn't have any issues in them so I might need to test a bit more to make sure

PR #1129, opened 2024-02-07
Preconsume archer-solid-cv4 and add fetch loop

callachennault:
This PR:
- Checks for additional problematic gene panel ARCHER-SOLID-CV4
- Adds loop to continue fetching queue until no problematic samples are found

### DIFF
--- a/import-scripts/preconsume_problematic_samples.sh
+++ b/import-scripts/preconsume_problematic_samples.sh
@@ -108,6 +108,7 @@ function attempt_to_consume_problematic_sample() {
     dmp_token="$1"
     sample_id="$2"
     type_of_problem="$3" # pass 'e' for event problems and 'm' for metadata problems
+    register_failures="$4"
     HTTP_STATUS=$(curl -sSL -w '%{http_code}' -o "$CONSUME_ATTEMPT_OUTPUT_FILEPATH" "${CVR_CONSUME_SAMPLE_URL_PREFIX}/${dmp_token}/${sample_id}")
     if [[ $HTTP_STATUS =~ ^2 ]] ; then
         if ! grep '"error": "' "$CONSUME_ATTEMPT_OUTPUT_FILEPATH" ; then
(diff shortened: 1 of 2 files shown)


import-scripts/preconsume_problematic_samples.sh
```diff
@@ -57,6 +58,7 @@ function set_cvr_fetch_url_prefix() {
 }
 
 function fetch_currently_queued_samples() {
+    FETCH_OUTPUT_FILEPATH="${FETCH_OUTPUT_FILEPATH}_${FETCH_NUM}"
```
callachennault, line 61: this part isn't working how I wanted - currently fixing

import-scripts/preconsume_problematic_samples.sh
```diff
@@ -168,20 +169,20 @@ function post_slack_message() {
...
-failed_to_consume_problematic_metadata_sample_list=()
-succeeded_to_consume_problematic_metadata_sample_list=()
-attempt_to_consume_problematic_samples
+while :
+do
+    consume_hardcoded_samples # temporary code
+    fetch_currently_queued_samples
+    detect_samples_with_problematic_events
+    detect_samples_with_problematic_metadata
+    exit_if_no_problems_detected
```
sheridancbio, line 183: This function currently does a sys.exit() system call, which would terminate the program immediately (before sending reports to slack or logging the problematic samples which were consumed.

I'd suggest turning this into a predicate function .. perhaps called "problems_were_detected" which returned 0 if there were problems and non-zero (1) if no problems were detected. Then this line could become:
    if ! problems_detected; then
        break
    fi

So the loop would end but we would continue to the log_actions and post_slack_message. However --- we probably also want to skip posting a slack message if there were absolutely no problematic samples consumed at all. So we may need a function to detect how many problematic samples actually were consumed (regardless of the number of times the loop was iterated).

Also .. we might need to check that at the end of attempting to consume problematic samples we blank out the files containing the sample ids which need to be consumed --- resetting them for the next loop iteration. I'm not sure if this is necessary, but it may be.

callachennault: good catch, I'll work on implementing that
to your last point, the files containing the sample ids to be consumed are rewritten on each loop iteration (currently the `consume_hardcoded_samples` function does this)

sheridancbio (approved):

PR #1132, opened 2024-02-12
Add script to generate age at seq for genie deliveries

callachennault:
Adds temporary script to generate `age_at_sequencing_reported_years` attribute (removed during MSK-CHORD rollout).  The attribute is needed for genie deliveries.

Need to discuss and implement long-term solution for attribute source


averyniceday (approved):

averyniceday (commented):
looks good, should probably discuss what's the long term plan if we decomission redcap

PR #1134, opened 2024-02-17
Do not assume consent status has changed if germline fetch fails

jamesqo:
Recently, the germline server at `draco.mskcc.org` has been returning an empty response for the consent values. This causes us to assume that all germline samples have been unconsented and remove them from the `mskimpact` mutations file, leading to https://github.com/knowledgesystems/pipelines-scrum/issues/1246.

This PR fixes that behavior by exiting early if no consent values are found.

/cc @averyniceday @callachennault

### DIFF
--- a/import-scripts/cvr_consent_status_checker.py
+++ b/import-scripts/cvr_consent_status_checker.py
@@ -41,6 +41,11 @@ def fetch_expected_consent_status_values():
     for field,url in CVR_CONSENT_STATUS_ENDPOINTS.items():
         response = urllib.urlopen(url)
         data = json.loads(response.read())
+
+        if data['status'] != "SUCCESS" or len(data['cases']) == 0:
+            print >> ERROR_FILE, "Could not retrieve %s consent status from germline server at %s. No action will be taken." % (field, url)
+            return None
+
         consent_values = {}
         for pt,status in data['cases'].items():
             if status:
@@ -192,6 +197,16 @@ def main():
         sys.exit(2)
 
     expected_consent_status_values = fetch_expected_consent_status_values()
+    if expected_consent_status_values is None:
+        print >> ERROR_FILE, "Unable to fetch consent status values, exiting..."
+        email_consent_status_report(
+            samples_to_requeue={},
+            samples_to_remove={},
+            gmail_username=gmail_username,
+            gmail_password=gmail_password
+        )
+        sys.exit(2)
+
     cvr_consent_status_fetcher_main(cvr_clinical_file, cvr_mutation_file, expected_consent_status_values, gmail_username, gmail_password)
 
 if __name__ == '__main__':
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ target
 **/src/main/resources/naaccr*.json
 **/repository.sqlite
 **/dependency-reduced-pom.xml
+.idea/
+.vscode/


import-scripts/cvr_consent_status_checker.py
```diff
@@ -192,6 +197,16 @@ def main():
         sys.exit(2)
 
     expected_consent_status_values = fetch_expected_consent_status_values()
+    if expected_consent_status_values is None:
+        print >> ERROR_FILE, "Unable to fetch consent status values, exiting..."
+        email_consent_status_report(
```
jamesqo, line 202: In its current state, the code will email something like:

```
Consent status update:


CONSENT STATUS GRANTED:

CONSENT STATUS REVOKED:
```

Wondering if I should make the message more specific

import-scripts/cvr_consent_status_checker.py
```diff
@@ -41,6 +41,11 @@ def fetch_expected_consent_status_values():
     for field,url in CVR_CONSENT_STATUS_ENDPOINTS.items():
         response = urllib.urlopen(url)
         data = json.loads(response.read())
+
+        if data['status'] != "SUCCESS" or len(data['cases']) == 0:
```
averyniceday, line 45: should also check for status code

averyniceday: also add a check (not ncessarily here) to see if total number of germline samples to remove is > than some threshold (can be arbitrary for now) - but above a certain threshold we send a warning and do not remove

averyniceday: don't think we should just return none -- we should return the original struct 
{ PARTC_CONSENT: {}, PARTA_CONSENT:{}}

And then run a check on it during the main and give alerts afterwards

import-scripts/cvr_consent_status_checker.py
```diff
@@ -192,6 +197,16 @@ def main():
         sys.exit(2)
 
     expected_consent_status_values = fetch_expected_consent_status_values()
+    if expected_consent_status_values is None:
```
averyniceday, line 200: None is too generic and doesn't cover the fact that PARTA might be working but PARTC might not - see above comment about preserving struct and then doing a check inside the main

averyniceday (changes requested):
good first try -- but we should preserve the original struct of what we were passing into the main and run checks on it there. 

Then based on status of our check we return a certain message -- can log the issue though as we run into them

averyniceday (changes requested):
can talk about it more in detail -- but believe this check is being run against the clinical sample file when really the check needs to be specifically for germline events in the mutation. 

Think the original code can be kept in place -- but maybe moving the check at the end of the remove_germline_samples() function as a validation before replacing the file.

import-scripts/cvr_consent_status_checker.py
```diff
@@ -118,10 +128,20 @@ def remove_germline_revoked_samples(cvr_mutation_file, revoked_germline_samples)
...
+                    num_removed_records += 1
+                    continue # Exclude this record from the new mutations file
             tmpfile.write(line)
     tmpfile.close()
+
+    pct_removed = 100*(num_removed_records / num_germline_records)
+    cutoff = 20 # If we're trying to remove too many records, then something's probably wrong with the server response. 20% is an arbitrary cutoff
+    if pct_removed >= cutoff:
+        print >> ERROR_FILE, "WARNING: %s%% of germline records had their Part C consent status changed. No action will be taken-- please double-check the response of the upstream server." % (pct_removed)
+        return
```
averyniceday, line 143: should also clean up the tmpfile

averyniceday (changes requested):
Check looks good - but should also include removal of tmpfile. WIll also need to propagate this info into the email notification we send out

import-scripts/cvr_consent_status_checker.py
```diff
@@ -141,16 +178,29 @@ def email_consent_status_report(samples_to_requeue, samples_to_remove, gmail_use
...
-            summary += '\n\t%s:\t%s samples' % (field, len(samples))
-            filename = field.lower() + '_consent_granted_report.txt'
-            generate_attachment(message, filename, samples)
+            missing_data = expected_consent_status_values[field] == {}
+            if missing_data:
+                part = "A" if field == PARTA_FIELD_NAME else "C"
+                summary += '\n\t%s:\tNo action. No response from Part %s server.' % (field, part)
+            else:
+                summary += '\n\t%s:\t%s samples' % (field, len(samples))
+                filename = field.lower() + '_consent_granted_report.txt'
```
averyniceday, line 187: do these two lines need to be outside the if/else block?

import-scripts/cvr_consent_status_checker.py
```diff
@@ -141,16 +178,29 @@ def email_consent_status_report(samples_to_requeue, samples_to_remove, gmail_use
...
+            missing_data = expected_consent_status_values[field] == {}
+            if missing_data:
+                part = "A" if field == PARTA_FIELD_NAME else "C"
+                summary += '\n\t%s:\tNo action. No response from Part %s server.' % (field, part)
+            elif field == PARTC_FIELD_NAME and not removed_germline_mutations:
+                # If too many samples had their Part C cosent status changed, there is probably an issue with the upstream server, so we didn't remove any records.
+                summary += '\n\t%s:\tConsent was revoked for an abnormally large number of samples-- no germline records were removed from the mutation file. Please double-check the response of the Part C server.' % (field)
+            else:
+                summary += '\n\t%s:\t%s samples' % (field, len(samples))
+                filename = field.lower() + '_consent_revoked_report.txt'
```
averyniceday, line 202: same here

import-scripts/cvr_consent_status_checker.py
```diff
@@ -141,16 +178,29 @@ def email_consent_status_report(samples_to_requeue, samples_to_remove, gmail_use
     if samples_to_requeue != {}:
         summary += '\n\nCONSENT GRANTED:'
         for field,samples in samples_to_requeue.items():
-            summary += '\n\t%s:\t%s samples' % (field, len(samples))
-            filename = field.lower() + '_consent_granted_report.txt'
-            generate_attachment(message, filename, samples)
+            missing_data = expected_consent_status_values[field] == {}
+            if missing_data:
+                part = "A" if field == PARTA_FIELD_NAME else "C"
```
averyniceday, line 183: can take out the part setting and just hardcode it into the summary message

PR #1176, opened 2024-04-04
disable bugged cvr requeue step for rollout

sheridancbio:
commented out requeue step to allow build/deploy of java21 cvr fetcher


sheridancbio:
No longer needed after #1183

PR #1186, opened 2024-05-16
Fix logic bug in CVR seq date backfill

jamesqo:
## Problem

CVR seq date file had blank seq dates for some samples. This was occurring with samples from the AZ / Sophia cohorts.

## Cause

Once a sample was in the CVR clinical sample file, its seq date would never get updated even if it was pulled again from the queue. This is because even though the queue contains the correct date, the value would get overwritten by a subsequent call to `processSeqDateFile` which backfills the old value from `seq_date.txt`.

## Solution

This PR updates the code so that the JSON seq date is pulled after the existing one. This way, the value is only backfilled for pre-existing samples that do not have an incoming update from the CVR queue.

/cc @averyniceday @callachennault @sheridancbio


callachennault (approved):
looks good to me @jamesqo - were you able to run a test CVR fetch with this change as a sanity-check?

PR #1188, opened 2024-05-31
Support for MSK Extract Projects import

n1zea144:


import-scripts/import-msk-extract-projects.sh
```diff
@@ -0,0 +1,63 @@
...
+
+source $PORTAL_HOME/scripts/clear-persistence-cache-shell-functions.sh
+
+IMPORT_FAIL=0
+extract_projects_notification_file=$(mktemp $tmp/import-msk-extract-projects-notification.$now.XXXXXX)
+# grab data from extract-projects repos
+fetch_updates_in_data_sources "extract-projects"
+
+# import data into msk portal
+$JAVA_BINARY -Xmx64g $JAVA_IMPORTER_ARGS --update-study-data --portal extract-projects-to-msk-portal --notification-file $extract_projects_notification_file --oncotree-version $ONCOTREE_VERSION_TO_USE --transcript-overrides-source mskcc --disable-redcap-export
```
sheridancbio, line 35: I think that ONCOTREE_VERSION_TO_USE must be defined somewhere in this script. Perhaps add a line after the setting of IMPORT_FAIL. For MSK studies (dmp) we use:
ONCOTREE_VERSION_TO_USE=oncotree_candidate_release
I'm not sure what expectations we have for other extract users, but they should probably fit in to the same (msk default) version of oncotree.

import-scripts/import-msk-extract-projects.sh
```diff
@@ -0,0 +1,63 @@
...
+if [ $? -gt 0 ]; then
+    echo "MSK Extract projects import failed!"
+    IMPORT_FAIL=1
+    EMAIL_BODY="Extract projects import failed"
+    echo -e "Sending email $EMAIL_BODY"
+    echo -e "$EMAIL_BODY" | mail -s "Update failure: MSK Extract Projects" $PIPELINES_EMAIL_LIST
+fi
+
+# get num studies updated
+if [[ $? -eq 0 && -f "$TMP_DIRECTORY/num_studies_updated.txt" ]]; then
```
sheridancbio, line 45: The first "$?" in this condition test will not have the exit status code from the attempted import on line 35. Instead it will have the exit stats of the "if" statement from 36 though 42.

If the intention is default num_studies_updated to zero any time the import exit status is an error, then I suggest using the IMPORT_FAIL variable which would be 0 if the import did not fail, and 1 if it did fail. So perhaps:
if [[ $IMPORT_FAIL -eq 0 && -f "$TMP_DIRECTORY/num_studies_updated.txt" ]]; then

import-scripts/import-msk-extract-projects.sh
```diff
@@ -0,0 +1,63 @@
...
+    num_studies_updated=`cat $TMP_DIRECTORY/num_studies_updated.txt`
+else
+    num_studies_updated=0
+fi
+
+# clear persistence cache
+if [[ $IMPORT_FAIL -eq 0 && $num_studies_updated -gt 0 ]]; then
+    echo "'$num_studies_updated' studies have been updated, clearing persistence cache for msk portals..."
+    if ! clearPersistenceCachesForMskPortals ; then
+        sendClearCacheFailureMessage msk update-msk-spectrum-cohort.sh
```
sheridancbio, line 55: change the name of the script reporting the error from update-msk-spectrum-cohort.sh to import-msk-extract-projects.sh

import-scripts/import-msk-extract-projects.sh
```diff
@@ -0,0 +1,63 @@
...
+    fi
+else
+    echo "No studies have been updated, not clearing persistence cache for msk portals..."
+fi
+
+# clean up msk-spectrum repo and send notification file
```
sheridancbio, line 61: change comment to refer to extract

import-scripts/import-msk-extract-projects.sh
```diff
@@ -0,0 +1,63 @@
...
+bash $PORTAL_HOME/scripts/datasource-repo-cleanup.sh $PORTAL_DATA_HOME/datahub_shahlab
```
sheridancbio, line 62: we want to clean up the repo for extract-projects, so this path should be changed from datahub_shahlab to the path to extract-projects.

import-scripts/import-msk-extract-projects.sh
```diff
@@ -0,0 +1,63 @@
...
+$JAVA_BINARY $JAVA_IMPORTER_ARGS --send-update-notification --portal extract-projects-to-msk-portal --notification-file "$extract_projects_notification_file"
```
sheridancbio, line 63: In order to email a notification for a new --portal argument, we currently need to code that into the pipelines codebase in order to support the new portal. Maybe it would be easier for you to simply change this call to the importer for sending the notification file into a hardcoded mail message compose/send. But I can show you the place in the pipelines codebase if you want to add properties to support the email sending via the importers (we will need to update the properties files and rebuild importers too)

sheridancbio (changes requested):
A few issues here. Most are just text changes, but we should talk about the email notification sending.

import-scripts/import-msk-extract-projects.sh
n1zea144: These are internal projects so I think using candidate release is acceptable.

sheridancbio (changes requested):
Everything looks ok except the construction of the notification email. To get the content of the notification email you can perhaps cat the header line (from a temp file) with the notification file and pipe that to the mail command.

import-scripts/import-msk-extract-projects.sh
```diff
@@ -0,0 +1,64 @@
...
+    if ! clearPersistenceCachesForMskPortals ; then
+        sendClearCacheFailureMessage msk import-msk-extract-projects.sh
+    fi
+else
+    echo "No studies have been updated, not clearing persistence cache for msk portals..."
+fi
+
+# clean up extract-projects repo and send notification file
+bash $PORTAL_HOME/scripts/datasource-repo-cleanup.sh $PORTAL_DATA_HOME/extract-projects
+echo -e "The following Extract projects have been added or update in the MSK cBioPortal:\n\n$extract_projects_notification_file" | mail -r "cbioportal-pipelines@cbioportal.org" -s "Updates to MSK cBioPortal (Extract Projects)" $PIPELINES_EMAIL_LIST
```
sheridancbio, line 64: I don't know where my initial comment went .. so restating here:
This looks like it will embed the notification filename into the email message (rather than the contents of the notification file). I suggest creating a new temp file and writing the header line ("The following Extract projects...
") into that temp file. The use:
$(cat $headertempfilename $notificationfilename) | mail ...

import-scripts/import-msk-extract-projects.sh
```diff
@@ -0,0 +1,65 @@
...
+    if ! clearPersistenceCachesForMskPortals ; then
+        sendClearCacheFailureMessage msk import-msk-extract-projects.sh
+    fi
+else
+    echo "No studies have been updated, not clearing persistence cache for msk portals..."
+fi
+
+# clean up extract-projects repo and send notification file
+bash $PORTAL_HOME/scripts/datasource-repo-cleanup.sh $PORTAL_DATA_HOME/extract-projects
+EMAIL_BODY=`cat $extract projects_notification_file`
```
sheridancbio, line 64: an underscore is needed between extract and projects_notification_file

import-scripts/import-msk-extract-projects.sh
```diff
@@ -0,0 +1,65 @@
...
+echo -e "The following Extract projects have been added or update in the MSK cBioPortal:\n\n$EMAIL_BODY" | mail -r "cbioportal-pipelines@cbioportal.org" -s "Updates to MSK cBioPortal (Extract Projects)" $PIPELINES_EMAIL_LIST
```
sheridancbio, line 65: Maybe change "added or update" to "added or updated"

sheridancbio (approved):
👍

PR #1190, opened 2024-06-03
Update import-msk-extract-projects.sh

sheridancbio:
grammar fix in extract message


sheridancbio:
making quick fix to grammar for new script from @n1zea144

PR #1193, opened 2024-06-25
fix wrong column index bux in oncokb-annotator

sheridancbio:
Fix https://github.mskcc.org/cdsi/oncokb-annotated-msk-impact/issues/1. column index for Mutation_Status in MAF has moved from 26 to 27 it seems.


zhx828 (approved):
LGTM, thanks!

PR #1194, opened 2024-07-25
Use Airflow REST API to trigger s3 pull DAG

callachennault:
This PR:

- Uses Airflow [REST API endpoint](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html#operation/post_dag_run) to trigger a DAG run of the cdm_etl_cbioportal_s3_pull DAG after the sample list has been uploaded to S3 by the cBioPortal ETL.

### DIFF
--- a/import-scripts/update-cdm-deliverable.sh
+++ b/import-scripts/update-cdm-deliverable.sh
@@ -64,10 +64,14 @@ AIRFLOW_CREDS=$(cat $AIRFLOW_ADMIN_CREDENTIALS_FILE)
 AIRFLOW_CERT=$PORTAL_HOME/pipelines-credentials/airflow-dev-cert.pem
 DAG_ID="cdm_etl_cbioportal_s3_pull"
 AIRFLOW_URL="https://airflow.cbioportal.dev.aws.mskcc.org"
+TMP_LOG_FILE="$PORTAL_HOME/tmp/trigger_s3_dag.log"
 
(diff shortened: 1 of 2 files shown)


import-scripts/update-cdm-deliverable.sh
```diff
@@ -58,3 +63,21 @@ rm $TMP_MERGED_SEQ_DATE
 rm $CDM_DELIVERABLE
 
 echo "`date`: CDM deliverable generation and upload complete"
+
+TMP_LOG_FILE="${PORTAL_HOME}/tmp/trigger_s3_dag.log"
```
averyniceday, line 67: maybe just for organization, move variables to top of the file?

averyniceday: and just a quick question -- why do we need the a tmp log file? Will it not get written to stdout and into the overnight pipelines log? 

Also, do we need to remove the tmp log file?

callachennault: I wanted to capture the HTTP status code and the contents of the HTTP response, and to get both I need to store the response contents somewhere (in a file). I know that this script writes to the fetch dmp log file, but I couldn't find that path stored in any var for me to use (I feel like the crontab controls the log path?) so I created the temp file. I can add a line to remove the file afterwards

averyniceday (approved):

PR #1216, opened 2025-01-08
Fix bug with sample filepath and store file in cdm repo

callachennault:
This PR fixes a bug introduced in https://github.com/knowledgesystems/cmo-pipelines/pull/1208 where an empty file is being uploaded to the S3 bucket. Also now stores a copy of the file that is uploaded to S3 in the CDM repo.


averyniceday (approved):
lgtm -- and just so I understand, we'll always have a persisted file on the server that's pushed every night?

callachennault:
> lgtm -- and just so I understand, we'll always have a persisted file on the server that's pushed every night?

yep! Previously we were only storing the sample file in s3, and not storing a copy of it on pipelines3. It's easier for me to debug issues by storing the file on both pipelines3 and on S3, so that I can isolate where potential data issues might be coming from. Or if, for example, an S3 upload error occurred, it's helpful to have a copy of the file that was being uploaded.

PR #1217, opened 2025-01-15
drop hgnc-1938 import support (defunct)

sheridancbio:
- drop non-existant proxy jxi2.mskcc.org
- fix PUBLIC_CLUSTER_KUBECONFIG path


callachennault (approved):
looks good to me @sheridancbio !

PR #1218, opened 2025-01-16
Modify subset-impact-data to include or exclude msk-chord elements based on flag

callachennault:
Changes in this PR:

- In `filter-clinical-arg-functions.sh`: Update to `filter_clinical_attribute_columns()` function to use a temp processing file. This change is propagated to function calls in `update-az-mskimpact.sh`, `update-sophia-mskimpact.sh`, and `update-cdm-deliverable.sh`
- In `filter-clinical-arg-function.sh`: New function `filter_cdm_clinical_elements()` that will remove CDM clinical patient and sample elements from a given study.
- In `subset-impact-data.sh`: Added flag `--include-msk-chord-data` to indicate whether CDM data should be included in the generated subset. If it should be excluded, `filter_cdm_clinical_elements()` is called to remove CDM clinical elements. If it should be included, CDM timeline files are subsetted via the `subset-cdm-timeline-files.sh` script.
- In `fetch-dmp-data-for-import.sh`: Flag `--include-msk-chord-data` is now passed to `subset-impact-data.sh` calls for appropriate studies (all subset studies except for sclc_mskimpact_2017).

### DIFF
--- a/import-scripts/filter-clinical-arg-functions.sh
+++ b/import-scripts/filter-clinical-arg-functions.sh
@@ -87,24 +87,48 @@ function find_clinical_attributes_to_filter_arg() {
 
 function filter_clinical_attribute_columns() {
     # Path to clinical file taken as an argument
-    clinical_attribute_filepath="$1"
+    clinical_filepath="$1"
 
     # Attributes to deliver in the clinical file
     attributes_to_deliver="$2"
 
-    # Path to output clinical file taken as an argument
-    output_filepath="$3"
+    # Temporary output file, removed at end of function
+    tmp_processing_file=$(mktemp -q)
 
     # Determine which columns to exclude in the patient file
-    if ! find_clinical_attributes_to_filter_arg "$clinical_attribute_filepath" "$attributes_to_deliver" ; then
+    if ! find_clinical_attributes_to_filter_arg "$clinical_filepath" "$attributes_to_deliver" ; then
         return 1
     fi
-    EXCLUDED_HEADER_FIELD_LIST="$clinical_attributes_to_filter_arg"
+    excluded_header_field_list="$clinical_attributes_to_filter_arg"
 
     # Filter out the columns we want to exclude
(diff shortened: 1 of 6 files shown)


import-scripts/filter-clinical-arg-functions.sh
```diff
@@ -87,24 +87,48 @@ function find_clinical_attributes_to_filter_arg() {
...
-    clinical_attribute_filepath="$1"
+    clinical_filepath="$1"
 
     # Attributes to deliver in the clinical file
     attributes_to_deliver="$2"
 
-    # Path to output clinical file taken as an argument
-    output_filepath="$3"
+    # Temporary output file, removed at end of function
+    tmp_processing_file=$(mktemp -q)
```
averyniceday, line 96: minor -- but wonder if there's a way to get this tmpfile name in the case proccess fails and we want to check what the output was

import-scripts/subset-impact-data.sh
```diff
@@ -29,31 +33,28 @@ case $i in
...
     ;;
+    -p=*|--portal-scripts-directory=*)
+    PORTAL_SCRIPTS_DIRECTORY="${i#*=}"
+    shift # past argument=value
+    ;;
     -m=*|--metadata-filename=*)
     METADATA_FILENAME="${i#*=}"
     shift # past argument=value
     ;;
+    -h=*|--include-msk-chord-data)
```
averyniceday, line 48: do we limit accepted values here? Might be good to check it's either false/true

averyniceday: is the value defaulted to false or true? I see it set to false on line 13

averyniceday (approved):
LGTM -- minor comment about getting the tmp file name in case we want to check intermediary

import-scripts/subset-impact-data.sh
callachennault: yep, its value defaults to false, and if the flag is included it is set to true

import-scripts/filter-clinical-arg-functions.sh
callachennault: just added a logging line to do this

PR #1219, opened 2025-01-17
change genie import process to support clickhouse

sheridancbio:
- use green/blue import process strategy

### DIFF
--- a/import-scripts/import-genie-data.sh
+++ b/import-scripts/import-genie-data.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+#------------ FLOCK USED FOR MUTUAL EXCLUSION ... SO SCRIPT WILL NOT START IF IT IS ALREADY RUNNING
 FLOCK_FILEPATH="/data/portal-cron/cron-lock/import-genie-data.lock"
 (
     # check lock so that script executions do not overlap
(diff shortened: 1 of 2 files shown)


import-scripts/import-genie-data.sh
```diff
@@ -58,67 +64,154 @@ FLOCK_FILEPATH="/data/portal-cron/cron-lock/import-genie-data.lock"
...
+            fi
+        fi
+        rm -f $CURRENT_DATABASE_OUTPUT_FILEPATH
+    fi
+#------------ DROP TABLES IN THE NON_PRODUCTION DATABASE TO MAKE SPACE FOR CLONING
+    if [ $SET_DATABASE_UPDATE_RUNNING_STATE_FAIL -eq 0 ] ; then
+        echo "we are going to clone $source_database_color into $destination_database_color and then import into $destination_database_color"
+        MYSQL_DATABASE_CLONE_FAIL=0
+        if ! $DROP_TABLES_FROM_MYSQL_DATABASE_SCRIPT_FILEPATH $MANAGE_DATABASE_TOOL_PROPERTIES_FILEPATH $destination_database_color ; then
+            #echo "Error during dropping of tables from mysql database $destination_database_color" >&2
```
callachennault, line 147: this commented line can probably be removed

callachennault (approved):
looks good to me @sheridancbio - my only comments/questions would be:
- should the other scripts called by the import-genie-data.sh script be included in this PR?
- there are some comments/TODOs in the `import-genie-data.sh` script that could probably be removed if they aren't going to be addressed (just for the sake of clarity). But if you want to keep them as reminders, that's also no problem
- the `download_clickhouse_sql_scripts_py3` script could be generalized to download any arbitrary list of files from an arbitrary git repo - it could be handy to have this kind of utility. But that's obviously not urgent and could be generalized in the future if needed

PR #1227, opened 2025-04-10
Git to s3 - DO NOT MERGE

mandawilson:


averyniceday (changes requested):
need to update the consume functions inside dmp-import-vars to consume from JSON stored in MSK_IMPACT_DATA_HOME instead of MSK_IMPACT_PRIVATE_DATA_HOME

averyniceday (approved):

PR #1228, opened 2025-04-15
Public import dag

callachennault:
This PR introduces two dags for importing to the public database: `import_public_dag.py` and `rollback_public_portal.py`. Bash wrapper scripts are also included.


sheridancbio (approved):
These scripts have essentially "worked" with some problem-addressing in the helper scripts. So looks good. Related to the portal-configuration repo PR I think we should think about ssh credental sharing. That may impact the ssh connection logic in this PR
But scripts are good. (see small recommendations)

PR #1235, opened 2025-07-18
Sophia seq date updates

callachennault:
Update Sophia cohort generation script to pull sequencing dates from `msk_impact_dates` schema


averyniceday (changes requested):
Split out functionality for volume vs table querying - better definition of "modes" 

Otherwise, general addition seems good to me!

averyniceday (approved):

PR #1248, opened 2025-10-22
Added Query Type to OncoKB MAF Annotator Script

jfkonecn:
# Introduction

This is a fix to get the TERT gene mutations annotated.

## Work Done

It looks like the OncoKB MAF annotator is picking the wrong query type for the MAF file.

Here is the [OncoKB annotator code](https://github.com/jfkonecn/oncokb-annotator/blob/09ff7e9614990747ff0f9132dad1b1e9825ac68d/AnnotatorCore.py#L448-L479) that tries to resolve the query type automatically. It searches for [these headers](https://github.com/jfkonecn/oncokb-annotator/blob/09ff7e9614990747ff0f9132dad1b1e9825ac68d/AnnotatorCore.py#L199-L206) to determine if the query type should be a `GENOMIC_CHANGE`. It looks like it's picking `HGVSP_SHORT` instead. Since most of the mutations do not have HGVSP_SHORT mutations, they end up not getting annotated.

It's unclear to me why the code stopped working after November 2023 (which is when Tomin's last known good file is dated at) since there doesn't appear to be any relevant code changes made past that date.

## Side Note

For this error: 
```sh
WARNING:AnnotatorCore:Unexpected reference genome, only GRCh37 and GRCh38 are supported. Skipping.`
```

`NCBI_Build` sometimes just has a number in it. For example instead of `GRCh37`, the column has a value of `37`.


sheridancbio (approved):
Hotfix deployed into production.

PR #1310, opened 2026-03-21
ClickHouse-only import triage DAG

jamesqo:


import-scripts/airflow-create-derived-tables.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
```
sheridancbio, line 1: Right .. this is the main benefit ... we don't need to sling we only need to re-create the derived tables

dags/import_clickhouse_base.py
```diff
@@ -0,0 +1,333 @@
...
+    )
+
+    with dag:
+        importer = config.importer
+        scripts_dir = config.scripts_dir
+        creds_dir = config.creds_dir
+        db_properties_filepath = f"{creds_dir}/{config.db_properties_filename}"
+        color_swap_config_filepath = f"{creds_dir}/{config.color_swap_config_filename}"
+        data_source_properties_filepath = f"{creds_dir}/{config.data_source_properties_filename}"
+        if len(config.target_nodes) != 1:
```
sheridancbio, line 118: 👍

dags/import_clickhouse_base.py
```diff
@@ -0,0 +1,333 @@
...
+                notification_filepath,
+            ),
+            "transfer_deployment": _script(
+                scripts_dir,
+                "airflow-transfer-deployment.sh",
+                scripts_dir,
+                db_properties_filepath,
+                color_swap_config_filepath,
+            ),
+            "clear_persistence_caches": _script(
```
sheridancbio, line 244: This is commented out above. I'm guessing that this is here as a placeholder in case we want a separate named step for clearing the persistence caches. The transfer_deployment script does clear persistence caches ... so maybe this will not be needed once we have all portals going through color swapping.

dags/import_clickhouse_base.py
```diff
@@ -0,0 +1,333 @@
...
+                db_properties_filepath,
+            ),
+            "create_derived_tables": _script(
+                scripts_dir,
+                "airflow-create-derived-tables.sh",
+                importer,
+                scripts_dir,
+                db_properties_filepath,
+            ),
+            "set_import_complete": _script(
```
sheridancbio, line 206: This is also commented out below --- because it is handled as part of transfer color.

dags/import_triage_clickhouse_dag.py
```diff
@@ -0,0 +1,98 @@
...
+                "msk-mind-datahub",
+                "pipelines-testing",
+                "genie",
+                "extract-projects",
+                "cmo-access"
+            ],
+        ),
+    },
+    wire_dependencies=_wire,
+    # Leave out the scheduled runs for now
```
sheridancbio, line 93: We can remove the nightly schedule if we grant DAG access to the cmo group which maintains source repo cmo-argos.

sheridancbio (commented):
Partial review - focus on the clickhouse dag definitions. Looks good.

import-scripts/airflow-clone-db.sh
```diff
@@ -33,8 +33,16 @@ fi
 
 SET_UPDATE_PROCESS_SCRIPT_FILEPATH="$PORTAL_SCRIPTS_DIRECTORY/set_update_process_state.sh"
 GET_DB_IN_PROD_SCRIPT_FILEPATH="$PORTAL_SCRIPTS_DIRECTORY/get_database_currently_in_production.sh"
-DROP_TABLES_FROM_MYSQL_DATABASE_SCRIPT_FILEPATH="$PORTAL_SCRIPTS_DIRECTORY/drop_tables_in_mysql_database.sh"
-CLONE_MYSQL_DATABASE_SCRIPT_FILEPATH="$PORTAL_SCRIPTS_DIRECTORY/clone_mysql_database.sh"
+
+if [[ "$PORTAL_DATABASE" == "triage-clickhouse" ]] ; then
```
sheridancbio, line 37: When this transition is complete and all imports are direct-to-Clickhouse we should come back and remove this conditional logic and the DB_TYPE specifier.

import-scripts/airflow-create-derived-tables.sh
```diff
@@ -0,0 +1,79 @@
...
+if [ ! -f $AUTOMATION_ENV_SCRIPT_FILEPATH ] ; then
+    echo "`date`: Unable to locate $AUTOMATION_ENV_SCRIPT_FILEPATH, exiting..."
+    exit 1
+fi
+source $AUTOMATION_ENV_SCRIPT_FILEPATH
+
+tmp=$PORTAL_HOME/tmp/import-cron-$PORTAL_DATABASE
+
+GET_DB_IN_PROD_SCRIPT_FILEPATH="$PORTAL_SCRIPTS_DIRECTORY/get_database_currently_in_production.sh"
+current_production_database_color=$(sh $GET_DB_IN_PROD_SCRIPT_FILEPATH $MANAGE_DATABASE_TOOL_PROPERTIES_FILEPATH)
```
sheridancbio, line 24: I believe the 'sh' inside the $( .. ) is unnecessary and possibly problematic. When bash is invoked with the command "bash" it has full functionality but when it is invoked with "sh" there are compatibility adjustments which effectively reduce it's functionality.

The script "get_database_currently_in_production.sh" has a first line like this:
"#!/usr/bin/env bash"

So executing the script directly will invoke the bash shell with full functionality.

Also --- I wonder whether we should be considering using get_database_currently_in_production_clickhouse.sh for the management database being in clickhouse?
Maybe this automatically gets called based on the properties in the config file

import-scripts/airflow-create-derived-tables.sh
```diff
@@ -0,0 +1,79 @@
...
+    clickhouse_schema_branch_name="public-portal-db-clickhouse-sql-for-import"
+fi
+if [ "$PORTAL_DATABASE" == "genie" ] ; then
+    clickhouse_schema_branch_name="genie-portal-db-clickhouse-sql-for-import"
+fi
+if [ "$PORTAL_DATABASE" == "msk" ] ; then
+    clickhouse_schema_branch_name="msk-portal-db-clickhouse-sql-for-import"
+fi
+if [ "$PORTAL_DATABASE" == "triage-clickhouse" ]; then
+    clickhouse_schema_branch_name="triage-portal-db-clickhouse-sql-for-import"
```
sheridancbio, line 67: I checked this branch in the cbioportal repo, and it points somewhere different than the msk/public/genie branches. We should reset this so that it points to the latest (1.0.9) derived table construction timepoint I think. The recent migration of those three databases did not include Triage, so I guess it was missed.

import-scripts/clear_cbioportal_persistence_cache.sh
```diff
@@ -126,6 +132,8 @@ portal_to_cache_database_number["genie-private-blue"]="3"
 portal_to_cache_database_number["genie-private-green"]="4"
 # pipelines3
 portal_to_cache_database_number["triage"]="1"
+portal_to_cache_database_number["triage-blue"]="10"
```
sheridancbio, line 135: Ok, I've added these two (10 and 11) to a spreadsheet I use to track these database integer number assignments.

If this rollout has already been completed, we could delete the lines in this file related to deployment "triage", which no longer exists.

sheridancbio (approved):
I've raised a couple of questions here (see below), and some adjustments are needed outside of this repo (see the branch reference for download of clickhouse.sql for the triage import). But I think this can be merged now and we can make adjustments in a future PR if necessary.

PR #1311, opened 2026-03-27
Remove unused DAGs

jamesqo:
Removes code for the triage / review / portals which are no longer in operation

Also removed msk branches for Airflow scripts -- msk doesn't run via Airflow

Not deployed to production yet


import-scripts/airflow-create-derived-tables.sh
```diff
@@ -60,9 +60,6 @@ fi
 if [ "$PORTAL_DATABASE" == "genie" ] ; then
     clickhouse_schema_branch_name="genie-portal-db-clickhouse-sql-for-import"
 fi
-if [ "$PORTAL_DATABASE" == "msk" ] ; then
```
sheridancbio, line 63: I'm not sure exactly what our plans are for getting the msk imports into airflow. I think we want to do that so that we do not rely on the crontab launch right?

Might we need this soon?

import-scripts/clear-persistence-cache-shell-functions.sh
```diff
@@ -49,16 +49,6 @@ function clearPersistenceCachesForMskGreenPortals() {
     fi
 }
 
-function clearPersistenceCachesForTriagePortals() {
-    all_triage_portals="triage"
-    clearPersistenceCachesForPortals "$all_triage_portals"
-}
-
-function clearPersistenceCachesForHgncPortals() {
```
sheridancbio, line 57: similarly, we will need to (eventually) make the hgnc portal live the next time the curation team is working on an update to the gene table. I guess we can add this back in when that time comes --- and it will need to be changed to a blue/green clickhouse approach anyway, so I guess deleting this now makes sense.

sheridancbio (approved):
Looks good. We will need to eventually convert msk and hgnc to a clickhouse-only import path and airflow DAG, but this mysql compatibility no longer is needed.

PR #1314, opened 2026-04-01
Make Import User Compatible with Pipelines3

averyniceday:
- Update smtp server to use msk relay 
- Update spf record in Squarespace
- SMTP server defined in file in pipelines-credentials, make script compatible with nodes both in/out of hospital network


import-scripts/importUsers.py
```diff
@@ -159,8 +162,11 @@ def send_mail(to, subject, body, gmail_username, gmail_password, sender=MESSAGE_
     for bcc_name in bcc:
         combined_to_list.append(bcc_name)
 
-    smtp = smtplib.SMTP_SSL(server, 465)
-    smtp.login(gmail_username, gmail_password)
+    if 'gmail' in server:
```
jamesqo, line 165: can you add a comment to explain the different port numbers / classes being used here?

jamesqo (approved):

PR #1315, opened 2026-04-02
Output empty strings for data_clinical_sample SO_COMMENTS, MSI_COMMENT, and data_sv Comments 

mandawilson:
Sensitive data has been sent in these fields so read them but output empty strings.  Note the objects are reading and storing the sensitive data, so if anyone uses the getter they could potentially output the sensitive data.

We blank them in the cvr fetch, and also in merge.py.

Fields that are now blank:

data_clinical_sample:
SO_COMMENTS
MSI_COMMENT

data_sv:
Comments


averyniceday (approved):
Looks good to me

PR #1316, opened 2026-04-03
Make sure all files are smudged when fetching data

jamesqo:
Fixes regression introduced by https://github.com/cBioPortal/datahub/pull/2274


sheridancbio (approved):
👍

sheridancbio:
Good detective work @jamesqo. When modifications which affect lfs behavior are made (such as in https://github.com/cBioPortal/datahub/pull/2274) it is important that the import operations and import scripting be considered and examined to see whether the changes are going to cause disruptions to our import pipelines.

PR #1317, opened 2026-04-08
ClickHouse version of import users script

jamesqo:
## Summary of changes

- Removed all read/write interactions with MySQL
- Moved all read operations with MySQL -> ClickHouse
- Replace usage of clickhouse CLI with clickhouse-connect library (will be useful in the future when this script is moved to the Airflow server)
- Update `users` / `authorities` table in a single batched SQL transaction
- Removed `--ssl-ca` flag which no longer applies (this was only relevant for RDS)

The portal.properties files also need to be updated in order to include clickhouse creds -- will submit a PR to pipelines-configuration later

cc @averyniceday @sheridancbio


jamesqo:
still need to test changes here

jamesqo:
Did some testing. Further changes:

- Converted script to Python3
- Added more logging
- Speed it up by doing a batched query for authorities in 1 swoop

import-scripts/importUsersClickhouse.py
```diff
@@ -233,58 +228,32 @@ def get_spreadsheet_title(client, ss):
...
-        sql_command_string = "INSERT INTO authorities VALUES %s;" % ",".join(values_string_items)
-        append_command_to_clickhouse_file(sql_command_string)
-    return emails_to_remove
+        print("new user: %s" % user.google_email, file=OUTPUT_FILE)
+        user_name = user.name
+        user_email = user.google_email.lower()
+        added_user_rows.append([user_email, user_name, user.enabled])
+        added_authority_rows += [[user_email, authority] for authority in user.authorities]
+    if added_user_rows:
+        ch_client.insert('users', added_user_rows, column_names=['email', 'name', 'enabled'])
```
averyniceday, line 249: let's add error checking like we were doing before for SQL here?

import-scripts/importUsers.py
```diff
@@ -478,49 +421,62 @@ def get_portal_properties(portal_properties_filename):
...
-                except MySQLdb.Error, msg:
-                        print >> ERROR_FILE, msg
+def update_user_authorities(spreadsheet, ch_client, sheet_records, portal_name):
+
+    # get map of current portal users
+    print('Getting list of current portal users from spreadsheet', file=OUTPUT_FILE)
+    all_user_map = get_new_user_map(spreadsheet, sheet_records, {}, portal_name)
+    if all_user_map is None:
+        return None
+    total = len(all_user_map)
```
averyniceday, line 458: keep this in a separate function?

averyniceday: and what does batch mean in this case? is this just all users?

import-scripts/importUsers.py
```diff
@@ -543,23 +499,22 @@ def add_rejected_users_to_worksheet(rejected_user_map, google_spreadsheet, clien
                 existing_emails.add(record[REJECTED_EMAIL_KEY].strip().lower())
     except Exception as e:
         # worksheet might not exist or be empty, start with empty set
-        print >> OUTPUT_FILE, 'Creating new rejected_users worksheet or worksheet is empty'
+        print('Creating new rejected_users worksheet or worksheet is empty', file=OUTPUT_FILE)
         existing_emails = set()
 
     # remove any rejected users that are already in the worksheet
-    for user_email in rejected_user_map.keys():
+    for user_email in list(rejected_user_map.keys()):
```
averyniceday, line 506: is there a reason this needs to be a list now? I see this referenced below without the change

import-scripts/importUsersClickhouse.py
```diff
@@ -478,49 +421,62 @@ def get_portal_properties(portal_properties_filename):
...
+                             parameters={'emails': emails})
+    db_authorities_map = {}
+    for row in result.result_rows:
+        db_authorities_map.setdefault(row[0].lower(), set()).add(row[1])
+
+    new_authority_pairs = []
+    for user in all_user_map.values():
+        sheet_authorities = set(user.authorities)
+        db_authorities = db_authorities_map.get(user.google_email, set())
+        added = [(user.google_email, authority) for authority in sheet_authorities - db_authorities]
```
averyniceday, line 471: is this authorities to add?

averyniceday (changes requested):
General changes look good - couple suggestions /questions regarding readiability and organization

import-scripts/import-portal-users.sh
```diff
@@ -29,32 +24,18 @@ MY_FLOCK_FILEPATH="/data/portal-cron/cron-lock/import-portal-users.lock"
...
+    $PYTHON3_BINARY $PORTAL_HOME/scripts/importUsersClickhouse.py --secrets-file $PIPELINES_CONFIG_HOME/google-docs/client_secrets.json --creds-file $PIPELINES_CONFIG_HOME/google-docs/creds.dat --properties-file $PIPELINES_CONFIG_HOME/properties/import-users/${GENIE_PRODUCTION_DATABASE_PROPERTIES_FILENAME} --send-email-confirm true --sender GENIE --gmail-username $GMAIL_USERNAME --gmail-password $GMAIL_PASSWORD --smtp-server $SMTP_SERVER >> "$USERSGENIELOGFILENAME" 2>&1
     CGDS_GENIE_IMPORT_STATUS=$?
-    clickhouse_commands_filepath=$(ls ${CLICKHOUSE_COMMANDS_PENDING_DIRPATH}/*.sql | head -n 1)
-    if [ $CGDS_GENIE_IMPORT_STATUS -eq 0 ] && [ -f "$clickhouse_commands_filepath" ] ; then
-        echo "executing clickhouse commands from $clickhouse_commands_filepath" >> "$USERSGENIELOGFILENAME"
-        date >> "$USERSGENIELOGFILENAME"
-        "$CLICKHOUSE_BINARY_FILEPATH" client --config-file="$GENIE_CLICKHOUSE_CLIENT_CONFIG_FILEPATH" --multiquery < "$clickhouse_commands_filepath"
-        CLICKHOUSE_IMPORT_STATUS=$?
```
sheridancbio, line 51: The logic here provided an inherent retry of failed database operations. It is ok for us to simplify this here, expecting clickhouse operations to always be successful on the first try, but if we encounter problems adding users to the clickhouse database we may need to return to a retry approach of some sort.

import-scripts/importUsers.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#!/usr/bin/env python3
```
sheridancbio, line 1: good -- we should be migrating all our python to a supported version

import-scripts/importUsersClickhouse.py
sheridancbio: Yes .. if the insert fails I wonder what happens. I suppose ch_client will throw an exception? At the very least we should confirm that if this happens we will get a stacktrace or some kind of indication in the run logs. A notice to our slack channel (letting us know that we failed an attempt to add a user to the user table) should probably also be made.

import-scripts/importUsers.py
```diff
@@ -686,65 +636,56 @@ def main():
             gmail_password = a
         elif o == '--properties-file':
             properties_filename = a
-        elif o == '--ssl-ca':
-            ssl_ca_filename = a
         elif o == '--send-email-confirm':
             send_email_confirm = a
         elif o == '--sender':
             sender = a
-        elif o == '--port':
```
sheridancbio, line 695: clickhouse services can run on different ports. I guess all that matters is that we can connect to the clickhouse.cloud service which is hosting the public genie databases, so if that works, then this is fine. But we might need to come back and adjust this if we choose different ports to configure (ssl / mysql compatibility / etc..)

sheridancbio (approved):
I breezed over the parts which seemed to be clear language conversions to accommodate python 3 syntax. So long as this works in production I approve. I suppose the inherent retry is still present because each time this runs we re-compute the list of users who are missing from the database and try to import them again.

import-scripts/importUsers.py
jamesqo: @averyniceday in python3 you're not allowed to modify a dictionary as you're iterating over it. we need to copy beforehand

import-scripts/importUsersClickhouse.py
jamesqo: i guess authorities that were added to the sheet

import-scripts/importUsers.py
jamesqo: resolved

import-scripts/import-portal-users.sh
jamesqo: I think this script will naturally retry because if it's not added to the database, we will keep picking up on that and trying to add them

jamesqo: but it's worth keeping tabs on in the logs

jamesqo:
@sheridancbio I'm wary of adding code for Slack notifications -- if something goes wrong with the network for example, we will be getting pinged every minute by this script. I did improve the error handling for failed DB operations however

import-scripts/importUsersClickhouse.py
jamesqo: I added back code that will email the failure message to users if there is an exception thrown during inserts.

jamesqo:
@copilot resolve the merge conflicts in this pull request

jamesqo:
We decided that we will be leaving the behavior on failure as-is for now but we should probably address it in the future so that users don't get spammed with emails if something goes wrong with our networking, for example.

import-scripts/import-portal-users.sh
```diff
@@ -13,11 +13,6 @@ MY_FLOCK_FILEPATH="/data/portal-cron/cron-lock/import-portal-users.lock"
     SMTP_SERVER=`cat $MAIL_SMTP_SERVER`
     GENIE_BLUE_DATABASE_PROPERTIES_FILENAME="portal.properties.genie.blue"
     GENIE_GREEN_DATABASE_PROPERTIES_FILENAME="portal.properties.genie.green"
-    GENIE_BLUE_CLICKHOUSE_CLIENT_CONFIG_FILEPATH="/data/portal-cron/pipelines-credentials/clickhouse_client_genie_blue_config.yaml"
```
jamesqo, line 16: removed these lines as we're no longer using the clickhouse CLI

averyniceday (changes requested):
Overall looks good to me - only one change -- let's just replace the contents of the existing importUsers.py script instead of having a separate importUsersClickhouse.py script. That way we do cleanup as part of this rollout, thanks!

sheridancbio:
Overall, I'm good with merging this PR. But one thing I'm not sure I understand ... why did we switch from using the command line executable "clickhouse client" and switch to using the python third party library "clickhouse_connect"?

I'm not against this ... but it means that now we have some parts of our code using `clickhouse client` and some parts using "clickhouse_connect". Both libraries come from the same organization ... but having 2 dependencies is worse than having 1. Perhaps we should stop using `clickhouse client` and do all clickhouse interactions through clickhouse_connect.

averyniceday (approved):

PR #1318, opened 2026-04-09
Script changes required for clickhouse-only public, GENIE, MSK

jamesqo:
- [x] New ClickHouse DAGs for clickhouse-only public/genie
- [x] Script changes required for new clickhouse importers
- [x] Move DAG execution from importer node to pipelines5
- [x] Deploy scripts / credentials / importer JARs to pipelines5
- [ ] Test full DAG flow (only doable once we have the production portals pointing to the the new clickhouse DBs)


sheridancbio (approved):
Looks good

import-scripts/import-cmo-data-msk.sh
```diff
@@ -61,8 +61,8 @@ FLOCK_FILEPATH="/data/portal-cron/cron-lock/import-cmo-data-msk.lock"
     fi
     DATA_SOURCE_MANAGER_SCRIPT_FILEPATH="$PORTAL_HOME/scripts/data_source_repo_clone_manager.sh"
     DATA_SOURCE_MANAGER_CONFIG_FILEPATH="$PORTAL_HOME/pipelines-credentials/importer-data-source-manager-config.yaml"
-    CMO_IMPORTER_JAR_FILENAME="/data/portal-cron/lib/msk-cmo-$destination_database_color-importer.jar"
-    CMO_JAVA_IMPORTER_ARGS="$JAVA_PROXY_ARGS $java_debug_args $JAVA_SSL_ARGS -Dspring.profiles.active=dbcp -Djava.io.tmpdir=$tmp -ea -cp $CMO_IMPORTER_JAR_FILENAME org.mskcc.cbio.importer.Admin"
+    CMO_IMPORTER_JAR_FILENAME="/data/portal-cron/lib/msk-clickhouse-importer-$destination_database_color.jar"
```
jamesqo, line 64: path to new clickhouse-only importer jarfiles

import-scripts/import-cmo-data-msk.sh
```diff
@@ -61,8 +61,8 @@ FLOCK_FILEPATH="/data/portal-cron/cron-lock/import-cmo-data-msk.lock"
...
+    CMO_JAVA_IMPORTER_ARGS="$JAVA_PROXY_ARGS $java_debug_args $JAVA_SSL_ARGS -Dspring.profiles.active=dbcp -Djava.io.tmpdir=$tmp -Dlog4j.appender.a.File=/data/portal-cron/logs/msk-cmo-clickhouse-importer.log -ea -cp $CMO_IMPORTER_JAR_FILENAME org.mskcc.cbio.importer.Admin"
```
jamesqo, line 65: pass in the log file explicitly instead of maintaining 4 separate jarfiles

import-scripts/update-msk-mind-cohort.sh
```diff
@@ -40,14 +40,15 @@ if [ "$destination_database_color" == "unset" ] ; then
     echo "Error during determination of the destination database color" >&2
```
jamesqo, line 40: I noticed that the invocation of this script in the fetch dmp wrapper script is commented out completely, what if I just delete this file?

import-scripts/set-data-source-environment-vars.sh
```diff
@@ -39,54 +39,3 @@ DATA_SOURCE_NAME_TO_FAILURE_LOG_MESSAGE[knowledge-systems-curated-studies]="cbio
 DATA_SOURCE_NAME_TO_EMAIL_BODY[knowledge-systems-curated-studies]="The cbio-portal-data data fetch failed."
 DATA_SOURCE_NAME_TO_EMAIL_SUBJECT[knowledge-systems-curated-studies]="Data fetch failure: cbio-portal-data"
 
-function fetch_updates_in_data_source {
```
jamesqo, line 42: To me this functionality appears to be supplanted by the data_source_repo_clone_manager script, went ahead and removed the function

jamesqo: Actually, I decided maybe it's better not to touch the business logic in here and make this a refactoring effort. Going to revert this change

dags/import_base.py
```diff
@@ -86,10 +86,13 @@ class ClickhouseImporterConfig:
...
 
 
-def _script(scripts_dir: str, script_name: str, *args: object) -> str:
+def _script(scripts_dir: str, script_name: str, *args: object, source_automation_env: bool = False) -> str:
     parts = [f"{scripts_dir}/{script_name}"]
     parts.extend(str(arg) for arg in args)
-    return " ".join(parts)
+    cmd = " ".join(parts)
+    if source_automation_env:
+        return f"source {scripts_dir}/automation-environment.sh && {cmd}"
```
sheridancbio, line 94: This copies the pattern we adopted for use with crontab. Basically we rely on the SSHOperator to execute a compound command (confusingly called a "pipeline" in bash syntax) on the target platform in the bash shell. This will work, but it means that everyone who adds a new task to a DAG in the future is going to have to adhere to the rule "you must always source automation-environment.sh before running your script". If we remember to do that, then we will be okay.
Another approach would be to keep all SSHOperator calls simple ... where they directly execute some script on the target machine **without** first invoking automation-environment.sh ... and then we require that all airflow task scripts begin by sourcing automation-environment.sh. This would have the advantage of making it easy to scan the airflow task scripts in cmo-pipelines and checking whether thay all source automation-environment.sh. If we want to check whether the compound commands in the airflow dags all include sourcing automation_environment.sh it might be more difficult ... or maybe not. I guess we could scan through the dags directory in this repo.

What I would strongly recommend is that we not source automation-enviornment.sh one way for some tasks and the other way for other tasks. Some of our scripts already include a sourcing of automation-environment.sh ... but if we are going to use the compound command approach on airflow it should be applied to **all** SSHOperator calls (even to scripts which already do the sourcing). I cannot tell whether that is the case with the PR as it stands now. Maybe @jamesqo can comment.

import-scripts/import-dmp-impact-data.sh
```diff
@@ -486,4 +486,4 @@ uploadToS3OrSendFailureMessage "$DMP_DATA_HOME" "" "mskimpact-databricks"
 # now remove the narrow format data cna file we just created
 rm "$MSK_SOLID_HEME_DATA_HOME/data_CNA_narrow.txt"
 
-exit 0
+exit $(( DB_VERSION_FAIL | IMPORT_FAIL_MSKSOLIDHEME | IMPORT_FAIL_ARCHER | IMPORT_FAIL_KINGS | IMPORT_FAIL_LEHIGH | IMPORT_FAIL_QUEENS | IMPORT_FAIL_MCI | IMPORT_FAIL_HARTFORD | IMPORT_FAIL_RALPHLAUREN | IMPORT_FAIL_RIKENGENESISJAPAN | IMPORT_FAIL_MSKIMPACT_PED | IMPORT_FAIL_SCLC_MSKIMPACT | IMPORT_FAIL_LYMPHOMA ))
```
averyniceday, line 492: This isn't necessarily a correct representation -- we may purposefully be skipping an import - let's not change the logic here.

import-scripts/test/test_msk_import_scripts.sh
```diff
@@ -0,0 +1,466 @@
...
+
+    # Create subdirs that scripts expect under MSK_DMP_TMPDIR etc.
+    mkdir -p \
+        "$MOCK_TMP/msk-dmp" \
+        "$MOCK_TMP/separate_working_directory_for_dmp" \
+        "$MOCK_TMP/import-cron-dmp-wrapper" \
+        "$MOCK_TMP/import-cron-cmo-msk"
+
+    # Create the checkpoint file so import-dmp-impact-data.sh doesn't loop
+    CHECKPOINT_FILEPATH="$MOCK_TMP/msk-import-script-go-ahead"
```
sheridancbio, line 82: This was put into the production scripts temporarily. The logic is : if the file does not exist, wait until the file does exist. This test is relying on this feature to be present in the production script. That is ok I suppose .. if we want to formalize this as a mechanism. I suppose all of this is temporary until we can containerize our importers. But maybe we should be asking whether this feature (the checkpoint file) is something we want to have present at all.

import-scripts/test/test_msk_import_scripts.sh
```diff
@@ -0,0 +1,466 @@
...
+    touch "$CHECKPOINT_FILEPATH"
+
+    # flock is Linux-only; provide a mock that always succeeds
+    mkdir -p "$MOCK_TMP/bin"
+    printf '#!/usr/bin/env bash\nexit 0\n' > "$MOCK_TMP/bin/flock"
+    chmod +x "$MOCK_TMP/bin/flock"
+    export PATH="$MOCK_TMP/bin:$PATH"
```
sheridancbio, line 89: rather than trying to shadow the original flock command using PATH manipulation, a less disruptive approach would be to do something like:
```
alias flock='/usr/bin/true'
```

import-scripts/test/test_msk_import_scripts.sh
```diff
@@ -0,0 +1,466 @@
...
+
+teardown() {
+    rm -rf "$MOCK_TMP"
+    rm -f "$MOCK_CALL_LOG"
+}
+
+# Touches all DMP study trigger files so every sub-import is attempted
+touch_all_dmp_triggers() {
+    touch \
+        "$MOCK_TMP/MSK_SOLID_HEME_IMPORT_TRIGGER" \
```
sheridancbio, line 162: This code (the test) will need to be maintained if we alter the set of studies we normally import. That's okay ... we will make updates here if we change the content of our import scripts.

import-scripts/fetch-and-import-dmp-data-wrapper.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-MY_FLOCK_FILEPATH="/data/portal-cron/cron-lock/fetch-and-import-dmp-data-wrapper.lock"
+MY_FLOCK_FILEPATH="${MY_FLOCK_FILEPATH:-/data/portal-cron/cron-lock/fetch-and-import-dmp-data-wrapper.lock}"
```
sheridancbio, line 3: I believe this change is for the creation of automated tests ... so that the automated tests can create a new FLOCK filepath and that new path will be used by the production script. However, the side effect of this is that any user who runs fetch-and-import-dmp-data-wrapper with a setting in "MY_FLOCK_FILEPATH" will actually use (in production) the filepath already present in MY_FLOCK_FILEPATH, which might cause the wrapper script to fail. To protect against this I would instead introduce a new variable .. perhaps FETCH_AND_IMPORT_DMP_DATA_WRAPPER_MOCK_FLOCK_FILEPATH. If that variable is set, then the value in that variable could be assigned to MY_FLOCK_FILEPATH .. otherwise MY_FLOCK_FILEPATH will be set to the default value.

But maybe all of this can be avoided just by doing the recommendation above (alias flock='/usr/bin/true')

import-scripts/fetch-and-import-dmp-data-wrapper.sh
```diff
@@ -72,38 +72,50 @@ function output_whether_preimport_steps_successfully_completed() {
...
-            /data/portal-cron/scripts/import-cmo-data-msk.sh
+            $PORTAL_HOME/scripts/import-cmo-data-msk.sh
+            if [ $? -ne 0 ] ; then IMPORT_FAIL=1 ; fi
             # Only run pdx updates on Friday->Saturday
-            if [ "$day_of_week_at_process_start" -eq 5 ] ; then
-                date
-                echo "executing import-pdx-data.sh"
-                /data/portal-cron/scripts/import-pdx-data.sh
-            fi
+            # if [ "$day_of_week_at_process_start" -eq 5 ] ; then
```
sheridancbio, line 93: If there isn't a "SKIP_PDX_PROCESSING" flag up top we could add one instead of checking in this hardcoded commenting out.

import-scripts/fetch-and-import-dmp-data-wrapper.sh
```diff
@@ -72,38 +72,50 @@ function output_whether_preimport_steps_successfully_completed() {
...
-            /data/portal-cron/scripts/import-msk-extract-projects.sh
-            #complete clickhouse update steps
-            $PORTAL_HOME/scripts/import-msk-postimport-steps-for-clickhouse.sh
+            # echo "executing update-msk-spectrum-cohort.sh"
+            # $PORTAL_HOME/scripts/update-msk-spectrum-cohort.sh
+            # if [ $? -ne 0 ] ; then IMPORT_FAIL=1 ; fi
+            # echo "executing import-msk-extract-projects.sh"
+            # $PORTAL_HOME/scripts/import-msk-extract-projects.sh
+            # if [ $? -ne 0 ] ; then IMPORT_FAIL=1 ; fi
+            if [ $IMPORT_FAIL -eq 0 ] ; then
```
sheridancbio, line 109: Here at the very end we are checking the status of IMPORT_FAIL. But why work hard trying to import cmo studies if the dmp imports already have shown failure. And why try importing mind if cmo imports have failed?  Any failure will cause us to discard the whole run, so as soon as there is one failure, skip over all remaining attempts to import (as well as skipping over the postimport steps)

import-scripts/fetch-ddp-and-import-cmo-access-data.sh
```diff
@@ -33,9 +33,23 @@ MY_FLOCK_FILEPATH="/data/portal-cron/cron-lock/fetch-ddp-and-import-cmo-access-d
     now=$(date "+%Y-%m-%d-%H-%M-%S")
     DATA_SOURCE_MANAGER_SCRIPT_FILEPATH="$PORTAL_HOME/scripts/data_source_repo_clone_manager.sh"
     DATA_SOURCE_MANAGER_CONFIG_FILEPATH="$PORTAL_HOME/pipelines-credentials/importer-data-source-manager-config.yaml"
-    IMPORTER_JAR_FILENAME="$PORTAL_HOME/lib/msk-cmo-blue-importer.jar"
+    GET_DB_IN_PROD_SCRIPT_FILEPATH="$PORTAL_HOME/scripts/get_database_currently_in_production.sh"
```
sheridancbio, line 36: We need to turn off this script now --- if it is still running. We cannot have two simultaneous import runs going into the msk portal database. So if this is running, we need to disable it. .. and we need to make sure that this does not try to run while the main (dmp wrapper) run is happening.  I guess we are getting some protection from the management database (which will show that an import is "running" and so we will exit) ... but we need to really think this through before we try to put this into production. We would not want a running instance of this script to block the dmp wrapper kickoff for instance ...

import-scripts/import-cmo-data-msk.sh
sheridancbio: This is good 👍

import-scripts/validate_blue_green_study.py
```diff
@@ -0,0 +1,240 @@
+#!/usr/bin/env python3
```
sheridancbio, line 1: the name of this script not descriptive enough, and is technically not accurate. We are not validating the study content ... we are detecting unexpected and dubious changes in study content.
... perhaps use "sanity_check_study_content_change.py"

import-scripts/validate_blue_green_study.py
```diff
@@ -0,0 +1,240 @@
...
+    with open(path) as f:
+        for line in f:
+            line = line.strip()
+            if not line or line.startswith('#'):
+                continue
+            if '=' in line:
+                key, _, value = line.partition('=')
+                props[key.strip()] = value.strip()
+    return props
+
```
sheridancbio, line 21: can we have just 1 blank line between functions?

import-scripts/validate_blue_green_study.py
```diff
@@ -0,0 +1,240 @@
...
+        source_cna = cna_event_count(source_db, source_cna_gp)
+        failed = dropped_too_much(dest_cna, source_cna)
+        report("CNA events", dest_cna, source_cna, failed=failed)
+        if failed:
+            failures.append(f"CNA event count dropped from {source_cna} to {dest_cna}")
+    else:
+        print(f"  {'CNA events:':<28} skipped (profile absent in one DB)")
+
+    # 10. Segment count
+    dest_seg = scalar(client, f"SELECT count(*) FROM {dest_db}.copy_number_seg WHERE cancer_study_id = {dest_id}")
```
sheridancbio, line 192: I've skimmed over this without probing too deeply. I don't expect us to encounter issues of this nature and these tests are a "belt and suspender" situation because if something drastically wrong gets put into production we should be able to roll back to the other color (yesterday).

sheridancbio (approved):
I'm approving this but I hope that some of my recommendations will be applied. Also, this should be squashed before (or during) merge.

sheridancbio:
Also .. before merging this we should make the CI tests pass

import-scripts/fetch-and-import-dmp-data-wrapper.sh
jamesqo: discussed -- unlikely scenario. Bash scopes variables to the file by default

import-scripts/fetch-and-import-dmp-data-wrapper.sh
jamesqo: if there are any failures we still want to import everything so we get notified of all errors at once

import-scripts/validate_blue_green_study.py
jamesqo: noted in comment in new PR

import-scripts/fetch-ddp-and-import-cmo-access-data.sh
jamesqo: added comment to new pr

dags/import_base.py
jamesqo: will source auto vars for each of the scripts

PR #1319, opened 2026-04-17
add script to set chosen environment variables

sheridancbio:
This is a workaround which will set chosen environment variables defined in automation-environment.sh

Why : some variables are needed for command line use, but when the automation-environment.sh is sourced from the .bash_profile login script, it winds up in an infinite loop because the automation-environment.sh script kicks off a subshell (which re-calls .bash_profile).

Using this script users can select the variables they want to set (perhaps automatically on login) or can create simple functions for setting the variables they want, such as with this (from the current .bash_profile):

```
function set_ssl_vars() {
    source /data/portal-cron/scripts/set-environment-vars-from-automation-script.sh SSL_CERT_FILE GIT_SSL_CAINFO
}
function set_auto_vars() {
    source /data/portal-cron/scripts/automation-environment.sh
}
echo "type 'set_ssl_vars' to enable use of clickhouse client and git at the command line"
echo "type 'set_auto_vars' to source the entire automation-environment.sh script"

```

### DIFF
--- a/import-scripts/set-environment-vars-from-automation-script.sh
+++ b/import-scripts/set-environment-vars-from-automation-script.sh
@@ -1,0 +1,98 @@
+#!/usr/bin/env bash
+
+# file scope variables
+THIS_SCRIPT_FILEPATH="/data/portal-cron/scripts/set-environment-vars-from-automation-script.sh"
+unset user_specified_parameter_names
+declare -a user_specified_parameter_names
+
+function script_was_directly_executed() {
+    [ "$0" == "$THIS_SCRIPT_FILEPATH" ]
+}
+
+function output_you_must_source_warning() {
+    echo "this script fails to function when you directly execute it as a new process."
+    echo "Instead you must use bash's 'source' or '.' to alter your current environment."
+    echo "try :"
+    echo "  source $THIS_SCRIPT_FILEPATH VARNAME1 VARNAME2 ... (or 'all')"
+}
+
+function set_user_selections() {
+    while [ $# -gt 0 ] ; do
+        user_specified_parameter_names+=("$1")
+        shift
+    done
+}
+
+function expression_contains_subshell() {
+    [[ "$1" == *\$\(* ]] || [[ "$1" == *\`* ]]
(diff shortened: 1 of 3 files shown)


import-scripts/set-environment-vars-from-automation-script.sh
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env bash
```
jamesqo, line 1: @sheridancbio I understand you wrote this script in order to implement `set_ssl_vars`. Do you think there's a scenario where we would only want those 2 variables, but not to grab the rest of the variables in automation-environment? Otherwise, I think this script could be somewhat complex to maintain.

jamesqo (dismissed):

import-scripts/set-environment-vars-from-automation-script.sh
sheridancbio: The script only sets a selected list of environment variables. The list of desired variables to set is passed as arguments by the caller, so it is easy to modify in the future. This script is not supposed to set any environment variable which the user did not request. (The user may request that all variables get set by calling the script with a single command line argument "all".)

sheridancbio: @jamesqo : I think maybe @averyniceday should try to answer this question. The two variables which are set in 'set_ssl_vars' are necessary for proper function of `git` and `clickhouse client` on pipelines5. I think we should not be defining the values for these variables in **both** automation-environment.sh and also in .bashrc/.bash_profile. So the script allows us to set just those two variables to enable "normal" interactive shell use of `git` and `clickhouse cloud` without getting the full set of environment setup which is present in automation_environment. My understanding is that @averyniceday wanted to be able to select manually (on startup) what the environment would look like --- either the full environment of the automated pipeline user (automation-environment.sh) or to only enable the self-signed certificate workaround.

But either way, I hope we can (before sprint planning tomorrow) finalize this choice, and merge this PR and move on.

jamesqo (approved):

PR #1324, opened 2026-04-27
retry git push on failure

sheridancbio:
This changes the `git commit --amend` into a plain `git commit`, and introduces a retry loop which re-attempt the push several times during the fetch-dmp-impact-data script processing.

We have observed repeated failures recently. This is an attempt to make it more likely to succeed with the git push operation. The retry loop is possibly a good strategy because almost all of the time of the push involves the upload of LFS objects, and presumably we will observe partial progress by uploading some of the necessary LFS objects on prior tries so there will be a shorter list of LFS objects to upload on future tries.

### DIFF
--- a/import-scripts/fetch-dmp-data-for-import.sh
+++ b/import-scripts/fetch-dmp-data-for-import.sh
@@ -1203,10 +1203,34 @@ MY_FLOCK_FILEPATH="/data/portal-cron/cron-lock/fetch-dmp-data-for-import.lock"
     # GIT PUSH
     printTimeStampedDataProcessingStepMessage "push of dmp data updates to git repository"
     # check updated data back into git
-    GIT_PUSH_FAIL=0
-    cd $DMP_DATA_HOME ; $GIT_BINARY add ./*; git commit --amend --no-edit; $GIT_BINARY push origin --force
-    if [ $? -gt 0 ] ; then
-        GIT_PUSH_FAIL=1
+    GIT_PUSH_FAIL=1
+    cd $DMP_DATA_HOME
+    MAX_PUSH_ATTEMPT_COUNT=13
+    remaining_push_try_count=$MAX_PUSH_ATTEMPT_COUNT
+    git_commit_message="DMP Fetch and Cohort Updates $(date +%Y_%m_%d)"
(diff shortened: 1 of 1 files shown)


import-scripts/fetch-dmp-data-for-import.sh
```diff
@@ -1203,10 +1203,34 @@ MY_FLOCK_FILEPATH="/data/portal-cron/cron-lock/fetch-dmp-data-for-import.lock"
...
     # check updated data back into git
-    GIT_PUSH_FAIL=0
-    cd $DMP_DATA_HOME ; $GIT_BINARY add ./*; git commit --amend --no-edit; $GIT_BINARY push origin --force
-    if [ $? -gt 0 ] ; then
-        GIT_PUSH_FAIL=1
+    GIT_PUSH_FAIL=1
+    cd $DMP_DATA_HOME
+    MAX_PUSH_ATTEMPT_COUNT=13
+    git_commit_message="DMP Fetch and Cohort Updates $(date +%Y_%m_%d)"
+    if $GIT_BINARY add ./* && $GIT_BINARY commit -m "$git_commit_message" --no-edit; then
```
callachennault: nitpick but we can remove the` --no-edit` flag since we're using `-m` for the commit message

import-scripts/fetch-dmp-data-for-import.sh
```diff
@@ -1203,10 +1203,34 @@ MY_FLOCK_FILEPATH="/data/portal-cron/cron-lock/fetch-dmp-data-for-import.lock"
...
+        remaining_push_try_count=$MAX_PUSH_ATTEMPT_COUNT
```
callachennault: initialize `remaining_push_try_count` before the if/else to ensure it has a known state if the git commit fails

callachennault (approved):

import-scripts/fetch-dmp-data-for-import.sh
sheridancbio: thanks .. I will drop it

import-scripts/fetch-dmp-data-for-import.sh
sheridancbio: 👍 fixing

PR #1338, opened 2026-06-11
Import CMO DAG

jamesqo:


sheridancbio (changes requested):
Code updates here seem ok, but a mutual exclusion mechanism is needed to prevent this script from running whenever the overnight dmp fetch/import cycle is running (and vice-versa). One thing which occurs in this script is the fetch of data source repositories (and the later cleanup), which could be corrupting (or at least disrupting) if it were to occur while the fetch-dmp-impact-data script is running.

Obviously simultaneous data import into the database must be prevented as well (with our current importer design).

Relying on the update management database state to provide mutual exclusion is not adequate, since much of the processing of the fetch-and-update-dmp-impact-data-wrapper script happens outside of the update status change (which is localized tot he import-dmp-impact-data script).

dags/import_base.py
```diff
@@ -192,6 +192,17 @@ def send_update_notification(notification_filepath: str, ssh_conn_id: str) -> No
                 db_properties_filepath,
                 color_swap_config_filepath,
             ),
+            "verify_import_not_in_progress": f"""
+                set_update_process_state_script="{scripts_dir}/set_update_process_state.sh"
+                manage_db_properties="{db_properties_filepath}"
+                if ! "$set_update_process_state_script" "$manage_db_properties" running ; then
+                    echo "Error: Another import process is currently running. The import cannot proceed while another import is in progress." >&2
```
sheridancbio, line 200: In theory, this error message might be generated for many reasons ... like a network outage or firewall problem, or the clickhouse database not responding fast enough.
To really know whether an update is running we could modify the set_update_process_state script here:
https://github.com/cBioPortal/cbioportal-core/blob/78ef2473cccbd8addfafbd060fe33a273157084c/scripts/clickhouse_import_support/set_update_process_state.sh#L159
and return a particular failure code for "already running so cannot set to running". Maybe we assign "8" to that condition for example. Then you could test for an exit status of "8" here. The other way (with the current script) would be to examine the stderr stream from running this command and look for "cannot set process status to running because current status is running" in the stream.

We also could add a script which simply reported the run state for a portal-db update.  Something which was the same as "get_database_currently_in_production.sh" but reported the state instead of the color.

This doesn't have to hold up this PR though. If it fails, it will still fail. We just may be a little confused by a message about something else running when actually we just have network congestion.

sheridancbio (changes requested):
This is good for stopping the airflow DAG from running while the overnight dmp wrapper is running. But we need the reverse too.

I've looked at the wrapper script ... the good news is that it sets the update management state to running before it starts the fetch script. So that is what we want. However, it currently prints a warning when its own call to set_update_process_state.sh fails because the state is already set to "running" and then continues with the import process. We need to instead abandon that night's run if the airflow cmo import run is still happening when the crontab kicks in and tries to launch the wrapper script.

See : https://github.com/knowledgesystems/cmo-pipelines/blob/b5651095b4946a57dc9873a4aab2fe66f2095b34/import-scripts/fetch-and-import-dmp-data-wrapper.sh#L59

This should be a pretty easy to add fix.  It also means that if we ever exit/crash and leave the management state for the msk portal as "running" we are going to be suspending our nightly runs until it is fixed manually. So we should post a loud message to slack channel #msk-pipeline-logs as well

import-scripts/fetch-and-import-dmp-data-wrapper.sh
```diff
@@ -55,16 +55,15 @@ function output_whether_preimport_steps_successfully_completed() {
...
-            echo "    and because this script will not run if a run is in progress, we are inferring that the prior run" 2>&1
-            echo "    is no longer running and the update management database is incorrect. However, this inference will" 2>&1
-            echo "    no longer be valid if we introduce any other processes which might independently run an update" 2>&1
-            echo "    into the msk portal database. The update management database has been reset by abandoning the attempt." 2>&1
+            echo "Error : the update management database shows an import is already in progress (status = 'running')." >&2
+            echo "    This could be the Airflow CMO MSK import DAG or a prior DMP wrapper run." >&2
+            echo "    Aborting this run to avoid concurrent data fetches and database imports." >&2
+            send_slack_message_to_channel "#msk-pipeline-logs" "string" \
+                "MSK portal nightly import skipped — another import is already in progress (management DB status is 'running'). You can re-run '$PORTAL_HOME/scripts/fetch-and-import-dmp-data-wrapper.sh' manually once the current import finishes."
```
sheridancbio, line 64: In order to re-run fetch-and-import-dmp-data-wrapper.sh manually, a user would need to source the automation-environment.sh script. But I think this is ok. If this ever does come up we should probably just let the normal cycle pick up the run for the following night (and not try a manual run).

sheridancbio (approved):

PR #1341, opened 2026-06-15
halt/fail dmp imports on a git push failure

sheridancbio:
If a git push does not complete successfully, we do not have a good basis for later on doing a git reset/purge to clean the repository followed by a s3 push (with delete) to clean up the s3 bucket as part of import-dmp-impact-data.sh


averyniceday (approved):
lgtm -- but could we add a comment in the code or somewhere in this PR that summarizes the state of the data and what we need to do to recover (if at all)

PR #1342, opened 2026-06-15
pull datahub for msk-automation-portal

sheridancbio:
do a git pull for datahub during cmo import operations

also restore call to script which updates imported studies (for groups etc)


averyniceday (changes requested):
not sure if I'm reading this right -- this looks like the opposite of what we want? PR looks to be adding back in the outdated updateCancerStudies.py (I think this updating groups to mysql) + removing datahub

PR #1344, opened 2026-06-19
Implement self-healing functionality for ArgoCD

jamesqo:


sheridancbio (approved):
The code changes here look good, but with self-heal there is now a chain of events which is not covered by the logic.

The trouble case is the following:
1. The script runs normally and reaches the point of changing the ingress rules to route traffic to DESTINATION_COLOR, which succeeds
1. The script calls set_update_process_state to "complete" the import process, changing the color to DESTINATION_COLOR in the management database
1. The script clears persistence caches
1. The script scales the replica count fully to DESTINATION_COLOR
1. The script adjusts the yaml files (deployments and ingress rules) in the local clone of knowledgesystems-k8s-deployments
1. The scripts attempts to commit and push the changes in the clone BUT FAILS (for whatever reason ... but perhaps because during the 15 minute period when the above was happening someone else pushed conflicting changes into the github repository)

If that happens, the failure to push will be reported in the logs, and then the script will exit. When the script exits, the trap will kick in and reset the argocd sync policy to perform self-heal. The cluster will be "healed" back to the prior configuration before starting the transfer deployment, which basically will switch the color away from DESTINATION_COLOR in both the ingress rules and the deployment replica counts.

I'm not sure how smooth that reversion will be --- we carefully transition and check status during the controlled transfer. Probably the self-heal will be more abrupt (with a minute or two of website outages).

Also, it will leave the cluster configured to the old color, while the update_management_database will be configured to the DESTINATION_COLOR as if the color transfer had succeeded. With the database out of sync with the cluster state, future imports will be blocked --- and we will have "dropped" the updates to the database which were attempted during the current import run.  We would need to manually examine the state and figure out why the database is out of sync with the cluster configuration, then make a judgement whether or not we should do a manual color transfer in order to "rescue" the import efforts which were in DESTINATION_COLOR. If not that, we would need to abandon the import efforts in DESTINATION_COLOR and notify the people who attempted those imports that they need to re-attempt them and finally manually correct the management database to match the cluster configuration.

It is true that this scenario is less likely to occur than the past problems we have experienced with people changing the cluster (or repo) without themselves syncing/committing the changes. But it will eventually happen (that someone makes a modification during the 15 minute color change). If the change by the other person in non-conflicting we are probably ok --- because we actually do a `git pull --rebase` just before the `git push`. But if a conflicting change is made, that will fail.

I've approved this PR, because maybe we are willing to tolerate the occasional occurrence of the problem case above. But perhaps we should get input from the team about whether we should automate a reset of the management database to the original color if we expect the self-heal to return the cluster to the original color.

PR #1364, opened 2026-08-04
Update msk-db-color-swap-config with beta mappings

dippindots:
Allow `beta.cbioportal.mskcc.org` to be updated as part of the deployment process


sheridancbio (approved):
👍

PR #1366, opened 2026-08-11
Containerized public import: Docker images, GH Actions builds, hackathon DAG

jamesqo:
Contains infra-only changes from the hackathon DAG work (docker images, DAGs, and GHA)

### DIFF
--- a/.github/workflows/build-core-image.yml
+++ b/.github/workflows/build-core-image.yml
@@ -1,0 +1,85 @@
+# Rebuilds the cbioportal-core importer image (docker/core.Dockerfile). The
+# build context is a cbioportal-core checkout while the Dockerfile comes from
+# this repo; both are checked out below and handed to buildx separately.
+#
+# Triggers:
+#   push to master  — when core.Dockerfile or this workflow changes
+#   repository_dispatch (cbioportal-core-push) — cbioportal-core CI can fire
+#     this with client_payload.core_branch to rebuild from a specific branch
+#   workflow_dispatch — manual runs; use core_branch to target a branch by name
+#     or core_ref to pin an exact SHA (core_ref takes precedence)
+name: Build cbioportal-core image
+
+on:
+  push:
+    branches: [master]
+    paths:
+      - docker/core.Dockerfile
+      - .github/workflows/build-core-image.yml
+  repository_dispatch:
+    types: [cbioportal-core-push]
+  workflow_dispatch:
+    inputs:
+      core_branch:
+        description: 'cbioportal-core branch to build from (default: main)'
+        required: false
+        default: 'main'
+        type: string
+      core_ref:
+        description: 'Exact cbioportal-core ref/SHA — overrides core_branch if set'
+        required: false
+        type: string
+
+env:
+  IMAGE: ghcr.io/cbioportal/containerized-importer-core
+  CORE_REPO: cBioPortal/cbioportal-core
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout cmo-pipelines (Dockerfile)
+        uses: actions/checkout@v4
+        with:
+          path: cmo-pipelines
+
+      - name: Resolve cbioportal-core ref
+        id: core
+        run: |
+          set -euo pipefail
+          ref="${{ inputs.core_ref || github.event.client_payload.core_ref }}"
+          if [ -z "$ref" ]; then
+            branch="${{ inputs.core_branch || github.event.client_payload.core_branch || 'main' }}"
+            ref=$(git ls-remote "https://github.com/${CORE_REPO}.git" "${branch}" | cut -f1)
+          fi
+          echo "ref=$ref" >> "$GITHUB_OUTPUT"
+
+      - name: Checkout cbioportal-core (build context)
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ env.CORE_REPO }}
+          ref: ${{ steps.core.outputs.ref }}
+          path: cbioportal-core
+
+      - name: Set up Docker Buildx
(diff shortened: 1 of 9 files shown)


.github/workflows/build-core-image.yml
```diff
@@ -0,0 +1,76 @@
...
+    inputs:
+      core_ref:
+        description: 'cbioportal-core ref to build (default: head of the tracked branch)'
+        required: false
+        type: string
+
+env:
+  IMAGE: ghcr.io/cbioportal/containerized-importer-core
+  CORE_REPO: cBioPortal/cbioportal-core
+  CORE_BRANCH: main
```
averyniceday, line 27: From here -- always building off main

docker/Dockerfile
```diff
@@ -0,0 +1,95 @@
...
+
+USER airflow
+COPY docker/requirements.txt /tmp/requirements.txt
+# --only-binary: fail loudly rather than silently compile from source
+RUN pip install --no-cache-dir --only-binary=:all: -r /tmp/requirements.txt \
+        --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-2.10.5/constraints-3.12.txt"
+
+# clickhouse import-support scripts from cbioportal-core, SHA-pinned; bump to advance:
+#   git ls-remote https://github.com/cBioPortal/cbioportal-core.git main
+ARG CBIOPORTAL_CORE_REF=6f69e5459a75b6b1dffdcb8c1d820d0426454577
```
averyniceday, line 67: from here - building off hash

dags/import_public_hackathon.py
```diff
@@ -0,0 +1,524 @@
...
+
+K8S_IMAGE            = "ghcr.io/cbioportal/containerized-importer-cmo:dev"
+K8S_IMAGE_VALIDATE   = "ghcr.io/cbioportal/containerized-importer-core:dev"
+VALIDATE_SCRIPT_PATH = "/scripts/importer/validateStudies.py"
+IMPORT_SCRIPT_PATH   = "/scripts/importer/metaImport.py"
+STUDY_LIST_VARIABLE_KEY = "available_study_ids"
+SCRIPTS_DIR = "/data/portal-cron/scripts"
+IMPORTER = "public"
+CREDS_DIR = "/data/portal-cron/pipelines-credentials"
+CREDS_SECRET_NAME = "import-credentials-test"
```
averyniceday, line 115: is this supposed to reference the secret created by create-pipelines-credentials-secret? Names don't match (import-credentials-test vs pipelines-credentials)

docker/create-pipelines-credentials-secret.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
averyniceday, line 1: where is this used - is this manually run from pipelines3? Is this better suited for k8s repo? need a way to keep SECRET_NAME in sync with your image build

docker/Dockerfile
```diff
@@ -0,0 +1,95 @@
...
+    chmod +x /usr/local/bin/yq; \
+    # saml2aws (extract all then find the binary — tarball layout varies by version)
+    curl -fsSL "https://github.com/Versent/saml2aws/releases/download/v${SAML2AWS_VERSION}/saml2aws_${SAML2AWS_VERSION}_linux_${GO_ARCH}.tar.gz" -o /tmp/saml2aws.tar.gz; \
+    mkdir -p /tmp/saml2aws-extract; \
+    tar -xzf /tmp/saml2aws.tar.gz -C /tmp/saml2aws-extract; \
+    find /tmp/saml2aws-extract -type f -name 'saml2aws' -exec install -m 0755 {} /usr/local/bin/saml2aws \;; \
+    rm -rf /tmp/saml2aws.tar.gz /tmp/saml2aws-extract
+
+# PORTAL_HOME tree the import-scripts assume; pipelines-credentials/ is a runtime mount
+RUN mkdir -p \
```
averyniceday, line 49: we shouldn't need cbio-portal-data? Some of these are remanants from the JAVA importer usage -- ideally meta_import does not depend on this, let's clean that up as part of this 

Thinking we probably need credentials, probably not git-repos, not lib (? depending how we organize)

averyniceday (changes requested):
core/importer docker images are not kept in sync - core.Dockerfile building off main, Dockerfile building off hardcoded hash

let's keep these in sync - and/or document a process that will keep these in sync

Other comments about further reducing image and versioning.

.github/workflows/build-core-image.yml
jamesqo: Will make this configurable

docker/Dockerfile
jamesqo: Will make this configurable

dags/import_public_hackathon.py
jamesqo: Yes, that's correct

.github/workflows/build-core-image.yml
jamesqo: Should be fixed

docker/Dockerfile
jamesqo: Removed git-repos, lib, and cbio-portal data. Don't want to risk removing anything else for the time being -- importers could still write to /logs, /tmp for example

jamesqo (commented):
replied to some comments

docker/create-pipelines-credentials-secret.sh
jamesqo: maybe it's just better to remove it from this PR. this is not part of the pipeline, it's more of a one-off script that I was keeping around for reference. we can just inspect the secret in the cluster in order to figure out what the schema is, should be easy enough to re-create the secret.

averyniceday (approved):

PR #1367, opened 2026-08-14
Requeue consent granted samples and run consent status checker script on all cohorts

callachennault:
Bug in `cvr_consent_status_checker.py`: Consent granted samples (where PARTA or PARTC consent status changes from 'NO' to 'YES') were never being requeued. The script now contains a function to run the `cvr_dmp_endpoint_utility.py` script as a subprocess to requeue the consent granted patients.

Bug in `fetch-dmp-data-for-import.sh`: The `cvr_consent_status_checker.py` script (which checks for consent granting and consent revocation) was only being run after the impact fetch, and was ignoring consent status changes in heme, access, and archer. Each cohort is now evaluated for consent status changes after the CVR fetch. Note that for impact and heme, the `cvr_consent_status_checker.py` occurs **after** the germline fetch so that freshly-written germline records are also caught and removed for any patient whose Part C consent was revoked.

### DIFF
--- a/import-scripts/cvr_consent_status_checker.py
+++ b/import-scripts/cvr_consent_status_checker.py
@@ -126,6 +126,11 @@ def cvr_consent_status_fetcher_main(cvr_clinical_file, cvr_mutation_file, expect
             record = dict(zip(header, map(str.strip, line.split('\t'))))
 
             for field in CVR_CONSENT_STATUS_ENDPOINTS.keys():
+                # If the API returned no data for this field, skip it entirely to avoid
+                # treating every patient as having revoked consent (the .get() default of 'NO'
+                # would otherwise cause all current-YES patients to appear as revoked).
+                if not expected_consent_status_values[field]:
+                    continue
                 current_consent_status = record[field]
                 expected_consent_status = expected_consent_status_values[field].get(record['PATIENT_ID'], 'NO')
                 # if current and expected values are the same then skip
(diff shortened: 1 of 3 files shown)


import-scripts/cvr_consent_status_checker.py
```diff
@@ -57,7 +59,45 @@ def fetch_expected_consent_status_values():
...
+    if not all_samples:
+        return
+
+    tmpfile_path = None
+    try:
+        fd, tmpfile_path = tempfile.mkstemp(prefix='consent_requeue_', suffix='.txt')
+        with os.fdopen(fd, 'w') as f:
+            f.write('\n'.join(sorted(all_samples)))
+
+        utility_script = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'cvr_dmp_endpoint_utility.py')
```
n1zea144: If PartC consent goes from NO to YES, do we have to use germline flag here?

n1zea144 (commented):
With the inclusion of other assays in the call to consent status check, possible other bugs in code that was not touched in this pr.  Something to check:

- on line 196 there is a possible division by 0 error - when a mutation file with zero germline records and any change in consent occurs (now reachable by hempact and access)
- on line 243 archer never runs germline removal, so any Part C revocation triggers a bogus "double-check the Part C server" warning since it doesn't pass a maf file.  remove_germline_revoked_samples should not get called/considered for Part C reporting when there is no maf.

callachennault:
> With the inclusion of other assays in the call to consent status check, possible other bugs in code that was not touched in this pr. Something to check:
> 
> * on line 196 there is a possible division by 0 error - when a mutation file with zero germline records and any change in consent occurs (now reachable by hempact and access)
> * on line 243 archer never runs germline removal, so any Part C revocation triggers a bogus "double-check the Part C server" warning since it doesn't pass a maf file.  remove_germline_revoked_samples should not get called/considered for Part C reporting when there is no maf.

@n1zea144  I pushed changes which address both of these comments. The  remove_germline_revoked_samples function was already not called when no MAF was provided but it was a bit ambiguous in the code so that's been made a bit clearer. The warning was being triggered separate from that function call. Both have now been fixed

import-scripts/cvr_consent_status_checker.py
callachennault: @n1zea144 This is a good point and might end up complicating things - 
If Part C goes from 'No' -> 'Yes', we need to requeue the sample on the somatic server so that 'Part C Consent' attribute is updated in the clinical file. But this does not add the germline variants back to the MAF. We would have to requeue the sample on the germline server, but the germline server requires normal sample format (`P00001-N01-IM3` instead of `P00001-T01-IM3`). This script is working with the tumor sample format, and as far as I'm aware we don't maintain a mapping of Tumor -> Normal sample IDs anywhere. We can either continue to assume that mPath is requeueing the germline samples when Part C consent is granted or we will have to add in some code to handle germline requeues (and maintain a mapping).

n1zea144 (approved):

PR #1382, opened 2026-09-03
Derive tables in standby database, not fixed secret database

jamesqo:
Two fixes to the `import_public_hackathon` DAG's `create_derived_tables_in_standby_database` task, needed before pointing the DAG at a real blue/green public database:

**1. Color-aware derive-tables target.** `rebuild_derived_tables.py` connects via `CLICKHOUSE_*` env vars, which were injected from the `hackathon-clickhouse-secret` k8s secret — a single fixed `database` value that ignores the blue/green standby color. Import writes the standby database (via `application.properties.{color}`), but derived tables were rebuilt in whatever database the secret named — wrong target half the time under real blue/green alternation. Now `_activate_standby_properties()` sets the `CLICKHOUSE_*` env vars from `manage_public_clickhouse_database_update_tools.properties` (`clickhouse_{blue,green}_database_name` selected by standby color) — the same config file the verify/clone/transfer tasks already use, so there is one source of truth. The `hackathon-clickhouse-secret` env injection is removed; the secret is no longer referenced by the DAG.

**2. Explicit derived-table SQL path.** The DAG stages the SQL at `/tmp/clickhouse.sql`, but `rebuild_derived_tables.py` falls back to `$PORTAL_HOME/populate_derived_tables.sql` when no path is given. Pass `--derived-table-sql /tmp/clickhouse.sql` explicitly to remove the filename coupling.

Known remaining caveat (unchanged by this PR): `clickhouse_server_additional_args` (`?secure=true...`) is not forwarded to the `clickhouse client` invocation; TLS behavior on native port 9440 is the same as before.

### DIFF
--- a/dags/import_public_hackathon.py
+++ b/dags/import_public_hackathon.py
@@ -124,6 +124,29 @@ CLICKHOUSE_CONFIG_FILE = f"{CREDS_DIR}/manage_public_clickhouse_database_update_
 S3_MOUNT_PATH = "/mnt/s3-data"
 S3_PVC_CLAIM_NAME = "databricks-s3-pvc"
 
+# Task IDs that may be listed in the skip_tasks param (dry-run support).
+SKIPPABLE_TASK_IDS = (
+    "clone_live_database_into_standby",
+    "import_into_standby_database",
+    "create_derived_tables_in_standby_database",
+    "transfer_deployment_color",
+)
+
+
+def _bash_skip_guard(task_id: str) -> str:
+    """Jinja prefix for a BashOperator command: exit 99 (BashOperator's skip exit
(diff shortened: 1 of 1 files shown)


jamesqo:
Added `study_prefix` param (45bdf407): reads studies from `s3://<bucket>/<prefix>/` instead of the root. Default `""` = unchanged behavior. Intended use: `{"database": "public", "study_prefix": "staging", "skip_tasks": ["transfer_deployment_color"], ...}` to dry-run modified studies against the real public standby without touching the root-level studies the scheduled import uses. `refresh_study_list` also stops listing `lfs/`, `embeddings/`, `staging/` as study ids.

